PR #13, opened 2020-10-28
Fix NameError on `typeprof --version`

ybiquitous:
When I run the command `typeprof --version`, the following error occurs:

```console
$ exe/typeprof --version
Traceback (most recent call last):
	1: from exe/typeprof:5:in `<main>'
/Users/masafumi.koba/git/ybiquitous/typeprof/lib/typeprof/cli.rb:78:in `parse': uninitialized constant TypeProf::CLI::VERSION (NameError)
```

This patch aims to fix the error. Here is an expectation:

```console
$ exe/typeprof --version
typeprof 0.3.0
```

### DIFF
--- a/lib/typeprof.rb
+++ b/lib/typeprof.rb
@@ -5,6 +5,7 @@ end
 
 module TypeProf end
 
+require_relative "typeprof/version"
 require_relative "typeprof/config"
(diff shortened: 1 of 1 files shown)


mame:
😅 Thanks!

PR #20, opened 2020-11-14
Add link to typeprof-playground

sanfrecce-osaka:
I think that typeprof-playground is very exciting because we can try typeprof on the Web.
But I think that we are a little bit difficult to notice typeprof-playground without knowing its name.
e.g. Searching "typeprof ruby" on Google, `Ruby Typeprof Playground` ranking is 7th.
So I added the link to it into README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,6 @@ end
 
 ## Playground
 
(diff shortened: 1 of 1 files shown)


mame:
Thank you!

PR #36, opened 2021-08-04
LSP: Support local variable definition jump

kateinoigakukun:
## Description

Added experimental support of `Go to local variable definition`. 

## Known Issue

This initial implementation doesn't fully support jumping into parameters because of the lack of convenient API.

## Demo

https://user-images.githubusercontent.com/11702759/128133125-afc80425-0a03-4472-bd94-d412beeb0b50.mov


mame:
Awesome, thank you!

PR #37, opened 2021-08-07
Should quote var name if identifier_keywords

ksss:
I think at least the identifier words in the variable names should be quoted.


mame:
Thank you!

PR #55, opened 2021-09-27
Add included dirs at the beginning of $LOAD_PATH

palkan:
My particular use-case is a bit tricky: I was trying to `typeprof` a [library which uses Ruby Next](https://github.com/palkan/anyway_config), which can not use `require_relative`, only _global_ `require`/`load`/etc. So, I wasn't able to run Typeprof against a local source code.

Another argument for this change is that now we can easily test both local and _gems_ version of the library by simply specifying (or not specifying) the `-I` option.


mame:
Hello! Thank you very much for your feedback.

I agree that adding given paths to the last of $LOAD_PATH was wrong.

This is a bit nitpicking, but here I'd like to use the same semantics of `-I` option of the ruby interpreter itself which adds given paths in turn:

```
$ ruby -I foo -I bar -e 'p $LOAD_PATH'
["/home/mame/work/ruby/foo", "/home/mame/work/ruby/bar", ...]
```

So, is this patch okay for you?

```diff
diff --git a/lib/typeprof/cli.rb b/lib/typeprof/cli.rb
index 7b4bd04..08e1ec0 100644
--- a/lib/typeprof/cli.rb
+++ b/lib/typeprof/cli.rb
@@ -24,6 +24,7 @@ module TypeProf
       show_version = false
       max_sec = max_iter = nil
       collection_path = RBS::Collection::Config::PATH
+      load_path = []

       opt.separator ""
       opt.separator "Options:"
@@ -31,7 +32,7 @@ module TypeProf
       opt.on("-q", "--quiet", "Do not display progress indicator") { options[:show_indicator] = false }
       opt.on("-v", "--verbose", "Alias to --show-errors") { options[:show_errors] = true }
       opt.on("--version", "Display typeprof version") { show_version = true }
-      opt.on("-I DIR", "Add DIR to the load/require path") {|v| $LOAD_PATH << v }
+      opt.on("-I DIR", "Add DIR to the load/require path") {|v| load_path << v }
       opt.on("-r FEATURE", "Require RBS of the FEATURE gem") {|v| gem_rbs_features << v }
       opt.on("--repo DIR", "Add DIR to the RBS repository") {|v| gem_repo_dirs << v }
       opt.on("--collection PATH", "File path of collection configuration") { |v| collection_path = v }
@@ -73,6 +74,8 @@ module TypeProf

       opt.parse!(argv)

+      $LOAD_PATH.concat(load_path)
+
       dir_filter ||= ConfigData::DEFAULT_DIR_FILTER
       rb_files = []
       rbs_files = []
```

palkan:
Should it be:

```diff
- $LOAD_PATH.concat(load_path)
+ $LOAD_PATH.unshift(*load_path) 
```

?

mame:
Ahhh, yes I meant it!

palkan:
Cool!
That should work perfectly (updated the PR).

PR #57, opened 2021-10-06
errors possibly be nil

mattn:


lib/typeprof/lsp.rb
```diff
@@ -330,7 +330,7 @@ def on_text_changed_analysis(res, definition_table, caller_table)
         end
 
         diagnostics = {}
-        res[:errors].each do |(file, code_range), msg|
+        res[:errors].to_a.each do |(file, code_range), msg|
           next unless file
```
mattn, line 334: code_range might have to be checked as nil.

mame:
Thank you, thank you! I guess the test failure is already fixed, so I've merged.

PR #67, opened 2021-10-16
Add `git clone typeprof`

znz:
Because `bundle install` failed without `../typeprof`.


YuheiNakasaka:
I've failed to run it by following the docs too. In addition to your patch, `git clone https://github.com/mame/rbswiki.rb` in line 6 also need to be fixed to `git clone https://github.com/mame/rbswiki` because of the repository not found.

mame:
I think there is no task for this issue, right? Closing.

PR #68, opened 2021-10-20
Support stdio transport for LSP

kateinoigakukun:
Currently, TypeProf has its own convention to tell TCP connection information, and communicate with VSCode extension using it.
 
But most LSP clients don't support such a convention, so users need to launch typeprof lsp server with fixed port and specify the port in vimrc.
See also: https://secret-garden.hatenablog.com/entry/2021/10/03/215203

On the other hand, stdio transport is well supported by many LSP clients, so this PR adds stdio transport support 

With this patch, you can integrate TypeProf for IDE with [vim-lsp](https://github.com/prabirshrestha/vim-lsp) easily.

```vim
autocmd User lsp_setup call lsp#register_server({
  \ 'name': 'typeprof',
  \ 'cmd': {server_info->['path/to/typeprof', '--lsp', '--stdio']},
  \ 'whitelist': ['ruby'],
  \ })
```

### DIFF
--- a/lib/typeprof/lsp.rb
+++ b/lib/typeprof/lsp.rb
@@ -7,6 +7,8 @@ module TypeProf
     if config.lsp_options[:stdio]
       reader = LSP::Reader.new($stdin)
       writer = LSP::Writer.new($stdout)
+      # pipe all builtin print output to stderr to avoid conflicting with lsp
+      $stdout = $stderr
       TypeProf::LSP::Server.new(config, reader, writer).run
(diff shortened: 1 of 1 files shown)


mame:
Hmm, thank you! I often use `Kernel#p` for debugging, so instead of rewriting `puts` to `STDERR.puts`, how about `$stdout = $stderr` (or `$stdout.reopen($stderr)`?) and keeping `puts` as is?

PR #69, opened 2021-11-10
Make Type.any, Type.nil, etc. be thread-local

mame:
Follow up of 0ad4c28e011ad85c822f6c680178d3c5e0f5d92d.

Now the Type objects are all thread-local.
"Type::Instance.new(<NilClass>)" returns a different object for each
thread. However, However, Type.nil cached the object in an instance
ariable of the class object, which brought inconsistency, i.e.,
"Type.nil === Type::Instance.new(<NilClass>)" may return false.

### DIFF
--- a/lib/typeprof/type.rb
+++ b/lib/typeprof/type.rb
@@ -422,22 +422,22 @@ module TypeProf
     end
 
     def self.any
-      @any ||= Any.new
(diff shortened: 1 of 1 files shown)


mame:
@CodingItWrong reported a bug of TypeProf-for-IDE that shows a false "undefined method" error, which was caused by this bug.

PR #72, opened 2021-12-09
Fix: Use cmd /c for windows

sevenc-nanashi:
LSP didn't work in windows:
```
[vscode] Try to start TypeProf for IDE
[vscode] typeprof is not supported for this folder: [object Object]
[vscode] because: Error: spawn bundle exec typeprof --version ENOENT
```
But using `cmd /c` worked for me, so I added this in typeprof LSP. If `process.platform` is `win32` and `shell` is undefined, LSP will use `cmd /c` like `bash -l -c`.


mame (approved):

mame:
Thank you! I cannot test it but I trust you. Will merge soon

PR #86, opened 2021-12-14
Failing Net::HTTP smoke test

wagenet:
Trying to evaluate net-http with its rbs fails. I'll try to narrow down the specific issue more if I can.


wagenet:
The issue specifically seems to be the `return yield(self)`.

wagenet:
Closed in favor of #87

PR #87, opened 2021-12-14
Resolve issue with yielding generics

wagenet:
I'm not sure that just commenting out the line is the correct thing to do, but it allows us to handle this scenario and doesn't appear to cause any other issues. This is important to handle since Net::HTTP's official types do something very similar.

I think ideally, we'd also infer the types for passthrough, but I can live with untyped for now.

### DIFF
--- a/lib/typeprof/analyzer.rb
+++ b/lib/typeprof/analyzer.rb
@@ -201,7 +201,7 @@ module TypeProf
       tys.each do |ty|
         raise "nil cannot be pushed to the stack" if ty.nil?
         ty.each_child do |ty|
-          raise if ty.is_a?(Type::Var)
+          #raise if ty.is_a?(Type::Var)
           #raise if ty.is_a?(Type::Instance) && ty.klass.type_params.size > 1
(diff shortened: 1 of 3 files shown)


mame:
Currently, I assume that the stack contains only concrete types. Pushing an abstract type breaks the assumption, which will bring many issues.

Instead of this PR, I'd like to remove type variables from the return value of `yield` by replacing them with `untyped`.

```diff
diff --git a/lib/typeprof/block.rb b/lib/typeprof/block.rb
index b890173..dc0b0f1 100644
--- a/lib/typeprof/block.rb
+++ b/lib/typeprof/block.rb
@@ -97,7 +97,8 @@ module TypeProf
       # check?
       #subst = { Type::Var.new(:self) => caller_env.static_env.recv_ty }
       # XXX: Update type vars
-      ctn[@ret_ty, caller_ep, caller_env]
+      ret_ty = @ret_ty.remove_type_vars
+      ctn[ret_ty, caller_ep, caller_env]
     end
   end
```

I'll commit this change with the test you created. Thank you very much!

PR #90, opened 2021-12-16
Don't generate duplicate class method definitions

wagenet:
I suspect we could probably share some code with `do_define_iseq_method` but I don't feel entirely confident in doing that correctly.

### DIFF
--- a/smoke/rbs-duplicate-class-method.rb
+++ b/smoke/rbs-duplicate-class-method.rb
@@ -1,0 +1,28 @@
+class Foo
+  def self.foo?
+    !!@foo
+  end
+
+  def self.bar?
+    !!@bar
+  end
+
+  def foo?
+    self.class.foo?
+  end
+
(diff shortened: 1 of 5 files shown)


smoke/rbs-duplicate-class-method.rb
```diff
@@ -0,0 +1,28 @@
...
+    self.class.bar?
+  end
+end
+__END__
+# Classes
+class Foo
+  self.@foo: bot
+  self.@bar: bot
+
+# def self.foo?: -> bool
```
wagenet, line 24: Previously, we'd get a second, uncommented `self.foo?` definition.

wagenet:
There's a related issue with Module self methods that I'm investigating.

mame:
Thank you!

> I suspect we could probably share some code with do_define_iseq_method

Agreed, I'll try doing it later

Issue #84, opened 2021-12-13
Methods defined in a Struct.new block end up on parent module


wagenet:
```rb
Customer = Struct.new(:name, :address) do
  def greeting
    "Hello #{name}!"
  end
end
```

Generated:

```rb
class Object
  private
  def greeting: -> String
end

class Customer < Struct[untyped]
  attr_accessor name(): untyped
  attr_accessor address(): untyped
end
```

PR #95, opened 2021-12-18
Failing test for struct with block

wagenet:
It looks like the path for structs is very different from Classes. Ideally we could share the path.

### DIFF
--- a/smoke/struct-with-block.rb
+++ b/smoke/struct-with-block.rb
@@ -1,0 +1,19 @@
+Foo = Struct.new(:foo) do
+  def foo?
+    !!foo
+  end
+
+  private
+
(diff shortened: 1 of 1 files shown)


mame:
Thanks. It's going to take some time to fix.

There is the same issue in `instance_eval` on a class object.

```
class Foo
end

Foo.instance_eval do
  def foo # should be defined as "Foo.foo", but currently it is handled as "Object#foo"
  end
end
```

mame:
#84

mame:
I'm reworking the algorithm on the v2 branch now. I copied this test as a known-issue. #192 I'd like to close this one, thanks!

PR #97, opened 2021-12-21
Fully qualify paths where necessary

wagenet:


smoke/rbs-base-namespace.rb
```diff
@@ -0,0 +1,14 @@
...
+  module Foo
+    class Bar < ::Foo::Bar
+    end
+  end
+end
+__END__
+# Classes
+module Qux
+  module Foo
+    class Bar < ::Foo::Bar
```
wagenet, line 11: Previously, this would be missing the leading `::`

mame:
Sorry for very late response! I've merged it.

I'm reworking the algorithm on the `v2` branch now. But your test in this issue doesn't work on v2 branch, so I copied your test as a known-issue. #192

PR #98, opened 2021-12-21
Handle superclass with identical name

wagenet:


smoke/rbs-matching-superclass.rbs
```diff
@@ -0,0 +1,7 @@
+class Foo
+end
+
+class Bar
+  class Foo < Foo
```
wagenet, line 5: Previously, this would look for `Bar::Foo` instead of `::Foo`.

PR #99, opened 2021-12-24
Suppress literal in void context warning

pocke:
`typeprof` command prints the following warning with `-w` Ruby option.

```bash
$ RUBYOPT=-w typeprof --version
(none):1: warning: possibly useless use of a literal in void context
typeprof 0.21.1
```

`RubyVM::AbstractSyntaxTree.parse("1")` causes the warning. So this PR changes the dummy code.


mame:
Sorry, I messed this PR up. I have committed your patch with you as a co-author at b11af89a596d3436befd607466c33749e1d6b06b. Forgive me for this.

PR #104, opened 2021-12-31
Fix for extension not working on fish shell

ankurp:
In fish shell `-l` option is not recognized and fails to start typeprof. This is a fix for fish shell users.


yoshitsugu:
@ankurp 
I've already fixed for fish shell in #102 . Could you try the latest master branch?

hsbt:
fish-3.5.1 have `-l` option same as `bash`.

```
-l or --login
       Act as if invoked as a login shell.
```

I'm not sure why it's not working.

PR #106, opened 2022-01-09
Update the version of typeprof described in the documentation

snaka:
When using the stable version of Ruby 3.1.0, some parts of the documentation were different from the actual behavior.

See also:
- https://github.com/mame/rbswiki/pull/1


mame:
Sorry for leaving this PR so long time. I have just merged it, thank you!

PR #110, opened 2022-02-21
Fix ExecuteCommand unresponsive

buzztaiki:
When I tried to executeCommand with emacs-lsp, it kept waiting for a response until it timed out.

This PR has been fixed to respond correctly.


mame:
Thanks!

PR #125, opened 2022-12-31
Use `name` property for more clearly log

ono-max:
# Before
```
typeprof is not supported for this folder: [object Object]
```
# After
```
typeprof is not supported for this folder: sampleProgram
```


mame (approved):

mame:
Thank you!

PR #126, opened 2022-12-31
Put only a command to run as a first argument in spawn method

ono-max:
See https://nodejs.org/docs/latest-v16.x/api/child_process.html#child_processspawncommand-args-options.


mame (approved):

mame:
I wonder if we should use something like [Shellwords.split](https://docs.ruby-lang.org/en/master/Shellwords.html#method-i-shellsplit) in Ruby. But your PR is better than the previous implementation, so I'll merge it. Thanks!

PR #127, opened 2022-12-31
Add tests for vscode extension

ono-max:
This PR cherry-picks the commit in https://github.com/ruby/typeprof/pull/126.

I confirmed these tests work in https://github.com/ono-max/typeprof/pull/4.


vscode/package.json
```diff
@@ -56,11 +56,17 @@
         "vscode:prepublish": "tsc -p ./",
         "compile": "tsc -watch -p ./",
         "postinstall": "node ./node_modules/vscode/bin/install",
-        "test": "node ./node_modules/vscode/bin/test",
+        "pretest": "tsc -p ./",
+        "test": "node ./out/src/test/runTest.js",
         "package": "vsce package"
     },
     "devDependencies": {
+        "@types/glob": "^8.0.0",
```
ono-max, line 64: dependencies libraries are based on https://github.com/microsoft/vscode-generator-code/blob/main/generators/app/templates/ext-command-ts/package.json#L34.

vscode/src/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
+import * as path from 'path';
```
ono-max, line 1: The codes in this file are borrowed from the generated code in https://github.com/microsoft/vscode-generator-code.

vscode/src/test/suite/index.ts
```diff
@@ -0,0 +1,40 @@
+import * as path from 'path';
```
ono-max, line 1: The codes in this file are borrowed almost from the generated code in https://github.com/microsoft/vscode-generator-code.

mame:
This is awesome! I knew I had to test this but I was lazy. Thank you very much!

PR #133, opened 2023-01-23
Check if the lock file exists correctly

ono-max:
Fix https://github.com/ruby/typeprof/issues/134.


mame (approved):

mame:
@ono-max I asked @pocke, who is the main author of rbs collection. According to him, the situation where there is rbs_collection.yaml and no rbs_collection.lock.yaml is exceptional; in such a situation, it would be good for typeprof to fail with the error message "Please execute 'rbs collection install'" or something. Could you please update your PR?

mame (approved):

PR #144, opened 2023-03-20
quiet output during tests for git commands

luke-gru:
Also, silence a warning for RubyVM::AbstractSyntaxTree.parse with keep_script_lines: literal used in void context


mame:
Sorry for my late response, and sorry I could no longer merge this PR cleanly. That's my fault. Instead, I have committed your change with you as a co-author at c4d0d8da3eb692133770ae9f1d3bff76cfd5b78d.

PR #146, opened 2023-03-31
Fix error with collection_path

ksss:
The problem occurred when using a combination of TypeProf and rbs collection.

```
TypeError: no implicit conversion of String into Integer
```

---

`collection_lock.gems` returns `Hash[String, library]`

https://github.com/ruby/rbs/blob/c24c65ce867c7d1cc3cc1ca871fc002579bd8854/sig/collection/config/lockfile.rbs#L48

And, `library` is record type.

https://github.com/ruby/rbs/blob/c24c65ce867c7d1cc3cc1ca871fc002579bd8854/sig/collection/config/lockfile.rbs#L24-L28

I wrote a test for using the rbs collection, but gave up because the `@builtin_env` class instance variable for `TypeProf::Config` is cached and would affect other tests.


ksss:
This issue resolved by https://github.com/ruby/typeprof/pull/149

PR #150, opened 2023-05-21
Remove no longer Goodcheck code

### LINKED ISSUE #148: test dependency sider/goodcheck is gone
While running the test for version 0.21.7 I ran into the issue that https://github.com/sider/goodcheck is gone. This makes https://github.com/ruby/typeprof/blob/v0.21.7/test/typeprof/goodcheck_test.rb unusable.

ybiquitous:
Closes #148 

(The test code has already removed via 0c7af5341a6317c8c50da6163a3b472d5f307476)

```console
$ git ls-files | grep -i goodcheck
(no match)

$ git grep -i goodcheck
(no match)
```


mame (approved):

mame:
Thank you!

PR #152, opened 2023-05-21
Fix TypeError for RBS::Collection::Config::Lockfile#gems

ybiquitous:
This patch aims to fix the following error:

```
/.../vendor/bundle/ruby/3.2.0/gems/typeprof-0.21.7/lib/typeprof/import.rb:34:in `[]': no implicit conversion of String into Integer (TypeError)

            collection_lock.gems.each {|gem| @loaded_gems << gem["name"] }
                                                                 ^^^^^^
```

A value (`library` type) of a Hash object returned by `RBS::Collection::Config::Lockfile#gems` should include the `:name` key:
- https://github.com/ruby/rbs/blob/ebb5257241c/sig/collection/config/lockfile.rbs#L48
- https://github.com/ruby/rbs/blob/ebb5257241c/sig/collection/config/lockfile.rbs#L24-L28


mame:
🤦‍♂️

mame (approved):

PR #158, opened 2024-03-15
Add support for future `putchilledstring` instruction

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/20205
Ref: https://github.com/ruby/ruby/pull/10262

Ruby 3.4 will very likely add a `putchilledstring` instruction which for what concerns typeprof is pretty much an alias of `putstring`.


mame:
I missed this PR. Merged, thanks!

casperisfine:
Thank you! ❤️

PR #159, opened 2024-03-18
Make the test suite compatible with `--enable-frozen-string-literal`

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/20205
Ref: https://github.com/voxpupuli/json-schema/pull/508

Since `typeprof` is tested as part of ruby-core CI, it needs to be compatible with the `--enable-frozen-string-literal` option.

cc @mame


mame:
Thanks!

PR #172, opened 2024-05-16
Add README.md for excited rubist

k-nasa:
As more people, like me, become interested in typeprof due to the presentation at RubyKaigi 2024, it's unfortunate that there's no README available. By adding a README, we hope that those who visit typeprof can work on good first issues.

ref: https://speakerdeck.com/mame/good-first-issues-of-typeprof


mame:
Thanks!

PR #173, opened 2024-05-16
Initial support lambda node

euglena1215:
Prior to this change, typeprof would crash on encountering code like:

```rb
def x = -> { 1 }
```

This was because typeprof did not yet support lambda. 
With this update, such lambda definitions are now recognized as `def x: -> Proc`. 
While the ideal scenario would be to infer the lambda's type directly (e.g., `^() -> Integer` as in RBS), this initial step addresses the crash issue and provides basic support.


mame:
Awesome.  I agree that it is reasonable as an initial step, thanks!

euglena1215:
I was able to work on this because I had a [GOOD FIRST ISSUE GUIDE](https://speakerdeck.com/mame/good-first-issues-of-typeprof). Thank you too!

PR #179, opened 2024-05-17
Support xstring

k-nasa:
I supported xstring_node and interpolated_xstring_node

```ruby
`echo foo`
`echo #{xstring_lit(10)}`
```


mame (approved):

mame:
👏

PR #180, opened 2024-05-18
Support retry

cobafan:
I supported retry

```
begin
rescue
  retry
end
```

### DIFF
--- a/lib/typeprof/core/ast/control.rb
+++ b/lib/typeprof/core/ast/control.rb
@@ -346,7 +346,7 @@ module TypeProf::Core
       end
 
       def install0(genv)
-        Source.new(genv.nil_type)
+        Source.new(Type::Bot.new(genv))
       end
     end
   end
--- a/lib/typeprof/core/ast/control.rb
+++ b/lib/typeprof/core/ast/control.rb
@@ -339,5 +339,15 @@ module TypeProf::Core
         ret
       end
     end
+
+    class RetryNode < Node
+      def initialize(raw_node, lenv)
+        super(raw_node, lenv)
+      end
+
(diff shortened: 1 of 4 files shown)


lib/typeprof/core/ast/control.rb
```diff
@@ -339,5 +339,15 @@ def install0(genv)
...
       end
     end
+
+    class RetryNode < Node
+      def initialize(raw_node, lenv)
+        super(raw_node, lenv)
+      end
+
+      def install0(genv)
+        Source.new(genv.nil_type)
```
mame, line 349: It would be better to return `Type::Bot.new(genv)` here instead of `genv.nil_type`. (Bot type means a type that does not return normally.)

mame:
First of all, thank you very much!

It requires more work to properly support `retry` because `retry` creates a loop:

```ruby
def foo
  n = 1
  begin
    raise if rand < 0.5
    # Here, n must be Integer | String
    n
  rescue
    n = "str"
    retry
  end
end
```

I don't think there are such many codes like this, though.

I am ok to merge this if you could add a scenario for this to scenario/known-issues. If you give it a try to support `retry` more properly, you may want to refer the implementation of `while` (but it may be a bit of a challenge).

cobafan:
Thank you for your kind review.

I haven't covered the pattern at this, but I think I need time to properly process it.
I hope to be able to continue the effort with other PR.

mame:
Would you like to get this PR merged now? If so, please press Ready for review. If you still want to do some work, I'll wait.

cobafan:
Yes, please.
Thank you very much🙇

mame:
Thank you!

PR #181, opened 2024-05-19
Add singleton class to known-issues

euglena1215:
Currently, we do not support `singleton_class_node`, which prevents us from handling syntax like the following:

```ruby
class Foo
  class << self
    def foo = :ok
  end
end
```

To address this as a known issue, I have added the corresponding test to known-issues.


mame:
Thanks!

PR #182, opened 2024-05-20
Support keyword arguments support partially

euglena1215:
Previously, methods defined with keyword arguments would have their keyword arguments ignored.

```rb
def foo(x:)
  x.to_i
end
```

→

```
def foo: -> Integer
```

With the added support for keyword arguments, keyword arguments can now be interpreted.
However, the following are not yet supported

- Method calls with keyword arguments
  keyword_hash_node because it does not support.
- Optional keyword arguments
  The scenario test `RuntimeError: AST diff does not work well; the completely same code generates a different AST` is annoying me and I gave up.

### DIFF
--- a/lib/typeprof/core/ast/method.rb
+++ b/lib/typeprof/core/ast/method.rb
@@ -70,20 +70,16 @@ module TypeProf::Core
       opt_keywords = []
       opt_keyword_defaults = []
 
-      kw = raw_args.keywords
-      if false
-      while kw
-        raise unless kw.type == :KW_ARG
-        lasgn, kw = kw.children
-        var, expr = lasgn.children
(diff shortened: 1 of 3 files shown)


euglena1215:
```
===============================================================================
Error: test: scenario/method/keywords.rb(ScenarioCompiler::ScenarioTest): RuntimeError: AST diff does not work well; the completely same code generates a different AST
scenario/method/keywords.rb:5:in `block (2 levels) in <class:ScenarioTest>'
scenario/method/keywords.rb:1:in `times'
scenario/method/keywords.rb:1:in `block in <class:ScenarioTest>'
===============================================================================
```

Hmmm, the test passed in tool/scenario_runner.rb, so I thought it was OK... I'll check later.

mame:
Thanks!

PR #186, opened 2024-05-22
Support RBS attribute definitions

akouryy:
This PR adds support for RBS's `attr_reader`, `attr_writer`, and `attr_accessor` type definitions. Their expected behavior in TypeProf appears to be the same as regular method `def`s, so I used `add_method_decl_box` for them.

Side notes:
- I believe the return type of `A#foo=` in the added scenario should be parenthesized as `A#foo= : (::Integer | ::Float) -> (::Integer | ::Float)`. If you agree, I can address this in a separate PR.
- This PR introduces calls with keyword arguments, which currently makes dogfooding impossible. I'm not entirely sure, but wrapping such pieces of code in `eval` might be a solution. Should I proceed with that?

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -300,6 +300,12 @@ module TypeProf::Core
       when RBS::AST::Members::Private
       when RBS::AST::Members::Alias
         SigAliasNode.new(raw_decl, lenv)
+      when RBS::AST::Members::AttrReader
+        SigAttrReaderNode.new(raw_decl, lenv)
+      when RBS::AST::Members::AttrWriter
+        SigAttrWriterNode.new(raw_decl, lenv)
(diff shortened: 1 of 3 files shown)


mame (approved):

mame:
I didn't realize I was missing those, thank you!

akouryy:
@mame Which do you prefer:
 `A#foo= : (::Integer | ::Float) -> ::Integer | ::Float` or
 `A#foo= : (::Integer | ::Float) -> (::Integer | ::Float)`?

mame:
Sorry, I skipped your notes.

* As for hover, it should be `(::Integer | ::Float)`.
* As for the keyword argument, yes, we should probably wrap it with "eval" for a while so that `ruby tool/dog_bench.rb` can go through it.

PR #193, opened 2024-05-25
Support match_last_line_node and interpolated_match_last_line_node

alpaca-tc:
In prism, conditional statements that are passed only RegExp is parsed as `MatchLastLineNode`/`InterpolatedMatchLastLineNode`. related: https://github.com/ruby/prism/pull/1442

Those nodes are now supported in this PR.

We almost never use the `$LAST_READ_LINE` + `if RegExp` feature, but in the process of editing the code, it may be parsed to those node.


mame:
Sorry for the delay! It looks good. I'll merge tonight or tomorrow. (I'm a bit busy until today :pray:)

mame (approved):

PR #194, opened 2024-05-27
Implement `SourceEncodingNode`

k-nasa:
Added support for SourceEncodingNode.

There was a problem with resolving classes included in the standard library, but I took an ad hoc solution for the time being.


mame (approved):

mame:
Thanks!

PR #195, opened 2024-05-27
Add a known issues scenario for argv

ShoheiMitani:
I added a scenario to the known issues because, when passing arguments via ARGV and explicitly converting them with to_s or to_i, they still remain untyped.

```rb
def foo(n)
  n.to_s
end

foo(ARGV[0].to_i)
```

### Test Results

```
Failure: test: scenario/known-issues/argv.rb(ScenarioCompiler::ScenarioTest)
scenario/known-issues/argv.rb:8:in 'block in <class:ScenarioTest>'
<"class Object\n" + "  def foo: (Integer) -> String\n" + "end"> expected but was
<"class Object\n" + "  def foo: (untyped) -> untyped\n" + "end">
```

### DIFF
--- a/scenario/known-issues/argv.rb
+++ b/scenario/known-issues/argv.rb
@@ -1,0 +1,11 @@
+## update
+def foo(n)
+  n.to_s
+end
+
+foo(ARGV[0].to_i)
+
+## assert
+class Object
+  def foo: (Integer) -> String
+end


mame (approved):

mame:
Thanks! https://github.com/ruby/typeprof/pull/199 will fix this issue.

PR #196, opened 2024-05-27
Implement `ClassVariableWriteNode` and `ClassVariableReadNode`

k-nasa:
I'm implementing `ClassVariableWriteNode` and `ClassVariableReadNode`.

I have implemented the base of the class variables, but I am having trouble passing the tests.

```ruby
## update
class B
  @@x = :ok

  def foo
    @@x
  end
end

## assert
class B
  def foo: -> :ok
end
```

I suspect that the scope control is not working properly, and the class variables cannot be found. Can you give me some advice?

### DIFF
--- a/lib/typeprof/core/env.rb
+++ b/lib/typeprof/core/env.rb
@@ -216,6 +216,12 @@ module TypeProf::Core
       mod.get_ivar(singleton, name)
     end
 
+    def resolve_cvar(cpath, singleton, name)
+      # TODO: include はあとで考える
+      mod = resolve_cpath(cpath)
+      mod.get_cvar(singleton, name)
+    end
+
     def resolve_type_alias(cpath, name)
       # TODO: include はあとで考える
       mod = resolve_cpath(cpath)
--- a/lib/typeprof/core/env.rb
+++ b/lib/typeprof/core/env.rb
@@ -219,7 +219,7 @@ module TypeProf::Core
     def resolve_cvar(cpath, singleton, name)
       # TODO: include はあとで考える
       mod = resolve_cpath(cpath)
-      mod.get_cvar(singleton, name)
+      mod.get_cvar(name)
     end
 
     def resolve_type_alias(cpath, name)
--- a/lib/typeprof/core/env.rb
+++ b/lib/typeprof/core/env.rb
@@ -216,7 +216,7 @@ module TypeProf::Core
       mod.get_ivar(singleton, name)
     end
 
-    def resolve_cvar(cpath, singleton, name)
+    def resolve_cvar(cpath, name)
       # TODO: include はあとで考える
       mod = resolve_cpath(cpath)
       mod.get_cvar(name)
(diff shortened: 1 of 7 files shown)


mame:
Wow, it's so well done.

There are two types of instance variables: those owned by the class and those owned by the instance.

```ruby
class Foo
  @var = 1 # The Foo class object has this ivar
  def foo
    @var = 1 # A Foo instance has this ivar, which is different from the above one
  end
end
```

`singleton` of ivars expresses this distinction.

There is no such distinction for class variables; both refer to the same class variable.

````ruby
class
  @@var = 1 # this and
  def fo
    @@var = 2 # this refer to the same class variable
  end
end
````

In other words, we don't need `singleton` for class variables at all. I think it would work if we just set `@cvars = {}` and eliminate singleton treatment elsewhere as well.

**Comment by k-nasa** [comment]
Thanks, Passed CI!

File: lib/typeprof/core/env/module_entity.rb [hunk 1]
```diff
@@ -350,6 +354,7 @@ def on_ancestors_updated(genv, base_mod)
         end
       end
       @ivar_reads.each {|ivar_read| genv.add_run(ivar_read) }
+      @cvar_reads.each {|cvar_read| genv.add_run(cvar_read) }
```
k-nasa, line 357: I'm just adding it without really understanding what it means, but is it ok?

mame: I think this change is good.

`on_ancestors_updated` is a hook that is called when the superclass of a class changes.
Consider the following example, and you remove the definition of `Foo::C`.

```ruby
class C
  @@foo = 1
end

module Foo
  # if this definition of `C` is removed,
  class C
    @@foo = "str"
  end

  class D < C # the meaning of `C` changes from `::D::C` to `::C`
    p @@foo # and this cvar also changes from `::D::C`'s `@@foo` to `::C`'s `@@foo`
  end
end
```

The superclass `C` in `class D < C` will change its meaning from `::D::C` to `::C`. In this case, we need to re-evaluate `@@foo`.

The code that you added enqueues all cvar reads in the class into the revaluation queue.

lib/typeprof/core/env.rb
```diff
@@ -216,6 +216,12 @@ def resolve_ivar(cpath, singleton, name)
       mod.get_ivar(singleton, name)
     end
 
+    def resolve_cvar(cpath, singleton, name)
```
mame, line 219: It would be better to make this method not accept singletons as well.

lib/typeprof/core/graph/change_set.rb
```diff
@@ -111,6 +111,12 @@ def add_ivar_read_box(genv, cpath, singleton, name)
       @new_boxes[key] = IVarReadBox.new(@node, genv, cpath, singleton, name)
     end
 
+    def add_cvar_read_box(genv, cpath, singleton, name)
```
mame, line 114: I don't think we need this `singleton` either.

mame (approved):

mame:
Thanks!

PR #200, opened 2024-05-29
Support `return x, y`

hoshinotsuyoshi:
Proposal: Support `return x, y`.

This change introduces the new class `DummyArrayNode`, which treats `return x, y` as `return [x, y]`.


mame:
Sorry for the delay! I was wondering if DummyArrayNode was a good design, but finally I think it's fine. Thanks!

PR #201, opened 2024-06-03
Implement class_variable_or_write_node

k-nasa:
SSIA

PR #202, opened 2024-06-03
implement or write

k-nasa:
SSIA


mame:
It conflicted with #201, so I tweaked it a bit and merged it. Thanks!

PR #204, opened 2024-06-03
Remove known-issues for class variable

k-nasa:
The class variable has been resolved, so it will be removed from known-issue.


mame (approved):

mame:
:clap::clap::clap: Thank you so much!

PR #205, opened 2024-06-03
Render positional arguments defined in RBS with hover

alpaca-tc:
The following are newly supported.

- Optional positional arguments
- Rest positional arguments
- Post positional arguments

NOTE: The keyword argument requires a change in `sig_type.rb`. I will change it in next PR.


mame (approved):

mame:
Thanks!

PR #206, opened 2024-06-03
Render keyword arguments defined in RBS with hover

alpaca-tc:
This is a follow-up PR to https://github.com/ruby/typeprof/pull/205

The following are newly supported.

- Required keyword arguments
- Optional keyword arguments
- Rest keyword arguments

---

Please give me your opinion for storing `req_keywords/opt_keywords` as `Hash`.

If passing `Hash` values to `#subnodes`, it raises an exception in `Node#each_subnode`. https://github.com/ruby/typeprof/blob/d9dd9b731f3d807885d3e47cf6280a7fac78e7ea/lib/ typeprof/core/ast/base.rb#L40
Therefore, passing Array that is fetched by`Hash#values` instead of Hash to `#subnodes`, but I am not confident in this implementation.


mame:
Thank you for your work!

> Please give me your opinion for storing req_keywords/opt_keywords as Hash.

Your current code looks nice enough, but maybe it would be better to have `Node#each_subnode` support a Hash.

By the way, this is not a problem with your patch, but I am concerned that handling required_keywords and optional_keywords separately does not preserve the order of keywords. For example, even if we write `def foo: (?opt_key: Integer, req_key: String)`, but TypeProf will pop up as `def foo: (req_key: String, ?opt_key: Integer)`. I wonder if the original order of keywords should be also saved.

alpaca-tc:
> Your current code looks nice enough, but maybe it would be better to have Node#each_subnode support a Hash.

@mame  Thank you for review! 💛
I just changed.  
https://github.com/ruby/typeprof/commit/4647c3b31697669ef46aff5e16c7df88bb0ccc32

> By the way, this is not a problem with your patch, but I am concerned that handling required_keywords and optional_keywords separately does not preserve the order of keywords. For example, even if we write def foo: (?opt_key: Integer, req_key: String), but TypeProf will pop up as def foo: (req_key: String, ?opt_key: Integer). I wonder if the original order of keywords should be also saved.

Current RBS doesn't support the original order of keywords. Steep also call this method.
https://github.com/ruby/rbs/blob/cb0af1d/lib/rbs/types.rb#L1127-L1140

IMO, I believe a consistent order is better for the development experience because `.rbs` is written by many people other than myself and the code style is not consistent.

mame (approved):

PR #207, opened 2024-06-04
Partial support for passing keyword arguments to CallNode

alpaca-tc:
Support KeywordHashNode and resolving basic keyword argument calls.

I have confirmed that it works in a simple case, but it is probably still incomplete.
However, it will be better than the current behavior where an exception is raised by a call to a keyword arguments.

### DIFF
--- a/lib/typeprof/core/ast/value.rb
+++ b/lib/typeprof/core/ast/value.rb
@@ -287,6 +287,46 @@ module TypeProf::Core
       end
     end
 
+    class KeywordHashNode < Node
+      def initialize(raw_node, lenv)
+        super(raw_node, lenv)
+        @keys = []
+        @vals = []
+        raw_node.elements.each do |raw_elem|
+          case raw_elem.type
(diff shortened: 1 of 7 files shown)


lib/typeprof/core/ast/value.rb
```diff
@@ -287,6 +287,47 @@ def install0(genv)
       end
     end
 
+    class KeywordHashNode < Node
```
alpaca-tc, line 290: This is the same behavior as `HashNode`, should I remove this and use `HashNode` instead of?

mame:
This is great! I was doing the same work, but you beated me.

lib/typeprof/core/ast/value.rb
mame: It's the exact same definition, right? Then, let's reuse HashNode for now. We can separate it if needed later

mame (approved):

mame:
Thank you!

PR #215, opened 2024-06-08
Support `next x, y`

hoshinotsuyoshi:
Similar to https://github.com/ruby/typeprof/pull/200, I thought it would be possible to implement the `next x, y` pattern using the DummyArrayNode added in https://github.com/ruby/typeprof/pull/200. How about that?


mame (approved):

mame:
Thanks!

PR #216, opened 2024-06-10
Support `$1` ~ `$9`

alpaca-tc:
Support for NumberedReferenceReadNode to handle `$1` ~ `$9`. Variables `$1` to `$9` are resolved to `String?`, but `$10` and beyond are resolved to `untyped`. According to RBS, only between `$1` and `$9` are currently defined.

For more details, refer to the RBS core global variables definition.
https://github.com/ruby/rbs/blob/ec5a819/core/global_variables.rbs#L77-L105

To support `$10` and beyond, we need to either add the definitions to RBS or modify `NumberedReferenceReadNode#install0` to return `Source.new(genv.str_type, genv.nil_type).`

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -149,6 +149,8 @@ module TypeProf::Core
         read = GlobalVariableReadNode.new(raw_node, lenv)
         rhs = AndNode.new(raw_node, read, raw_node.value, lenv)
         GlobalVariableWriteNode.new(raw_node, rhs, lenv)
+      when :numbered_reference_read_node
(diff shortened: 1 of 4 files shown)


mame (approved):

mame:
Thanks!

BTW, I would like to make `$1` return a `String`, not `String?`.  `String?` is indeed "correct", but I think it is often inconvenient because it brings more false warnings than finding actual errors.
This is because almost always when I write `$1`, I expect it to return a String (unless it is written as a condition, like `if $1`).
So I think it would be good to make `$<number>` always return String.

PR #217, opened 2024-06-10
Fix an error when parsing `case` statement without predicate

alpaca-tc:
I found that a `case` statement without a predicate throws an error, so I fixed it.


mame (approved):

mame:
Thanks!

PR #218, opened 2024-06-10
fix: interpolated_string_node

alpaca-tc:
Fixed a bug that couldn't parse the following string.

```
"a"\
  "#{do_something}"
```


mame:
Good catch!

PR #220, opened 2024-06-11
$1 returns String

alpaca-tc:
follow-up #216 

related: https://github.com/ruby/typeprof/pull/216#issuecomment-2160138592

> @mame
> BTW, I would like to make $1 return a String, not String?. String? is indeed "correct", but I think it is often inconvenient because it brings more false warnings than finding actual errors.
> This is because almost always when I write $1, I expect it to return a String (unless it is written as a condition, like if $1).
> So I think it would be good to make $<number> always return String.


mame (approved):

mame:
Thanks!

PR #221, opened 2024-06-16
Add `elsif` to known-issues

alpaca-tc:
`elsif` + block + local variable assignment causes the following exception.

```
Error: test: scenario/known-issues/elsif_and_block.rb(ScenarioCompiler::ScenarioTest): NoMethodError: undefined method `show_name' for an instance of TypeProf::Core::Source
.../typeprof/lib/typeprof/core/ast/call.rb:104:in `block in install0'
```

I have confirmed that the error does not occur if the `call.rb` is modified as follows, but I have added it to known-issues because I think this problem is caused by unsupported `elsif` behavior.

```
diff --git a/lib/typeprof/core/ast/call.rb b/lib/typeprof/core/ast/call.rb
index 8b2c4e9..f6a76f4 100644
--- a/lib/typeprof/core/ast/call.rb
+++ b/lib/typeprof/core/ast/call.rb
@@ -101,7 +101,7 @@ module TypeProf::Core
           vars.uniq!
           vars.each do |var|
             vtx = @lenv.get_var(var)
-            nvtx = vtx.new_vertex(genv, "#{ vtx.show_name }'", self)
+            nvtx = vtx.new_vertex(genv, "#{ vtx.is_a?(Vertex) ? vtx.show_name : "???" }'", self)
             @lenv.set_var(var, nvtx)
             @block_body.lenv.set_var(var, nvtx)
           end
```

### DIFF
--- a/scenario/known-issues/elsif_and_block.rb
+++ b/scenario/known-issues/elsif_and_block.rb
@@ -1,0 +1,19 @@
+## update
+def elsif_and_block
+  if true
+  elsif true
+  else
+    w = 0
+  end
+
+  yield_self do
+    w = 0
+  end
+
+  1
+end
+
+## assert
+class Object
+  def elsif_and_block: -> Integer
+end


mame:
Sorry! I completely missed your two PRs. Handling `elsif` was completely wrong. #224 will fix the issue.

As for the vertex name, it was useful for debugging in the early stages of development. But after the analyzer is complicated, it is not so useful, so I am thinking of removing them.

PR #222, opened 2024-06-17
Added support for SingletonClassNode

alpaca-tc:
Add support for the following code.

```ruby
class Foo
  class << self
    def foo
      puts 'foo'
    end
  end
end
```

## Additional information

- Added `LocalEnv#implicit_receiver`.
  - This is used when the receiver of `def` is implicitly `self` in a syntax like `class << self`.
  - I feel this is not a good implementation, but I could not find a best implementation. Please review it.
    - Better to allow `CRef` to represent singleton_class...? hmmm... I can't find a best solution 😭 
- I've changed a `Core::Service#dump_declarations` to keep nested class/module nodes in the stack.
  - This allows `Node#traverse` to control the output of `SingletonClassNode` without duplicating the class definition if it already outputs `ClassNode/ModuleNode`.
- Added known-issues for syntax that could not be handled.

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -13,7 +13,7 @@ module TypeProf::Core
       Fiber[:comments] = result.comments
 
       cref = CRef::Toplevel
-      lenv = LocalEnv.new(path, cref, {}, [])
+      lenv = LocalEnv.new(path, cref, {}, [], nil)
 
       ProgramNode.new(raw_scope, lenv)
     end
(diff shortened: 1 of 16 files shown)


mame:
Well, I am unsure if extending LocalEnv is good or not. Let me consider it more.
RBS only deal with class and instance methods, so it is fine to ignore `class << self; def self.foo` (we may want to show a warning). `class << non_class_const` is also fine to ignore with a warning.

mame:
Sorry for the late action. I am still not sure what the best solution is, but I tried to make `CRef` have scope_level, instead of LocalEnv. Now `self` in cpath is handled uniformly by `AST.parse_cpath`. I'll merge it with this for now, though the warnings are not implemented yet.

PR #223, opened 2024-06-21
Fix the handling of multibyte characters

NotFounds:
When a Ruby file contains multibyte characters such as Japanese, Chinese, or emojis, features like "go to definition" or "hover" don't work correctly. Because multibyte characters are not considered in position calculations.

Prism provides a code unit API to address these issues. My understanding, while the position calculation cost increases when multibyte characters are present, there is little to no change in cost otherwise.
https://docs.ruby-lang.org/en/master/Prism/Location.html#method-i-start_code_units_column

In this pull request, I fix the code to use the code unit API to correctly calculate positions with multibyte characters.

|Before|After|
|------|-----|
|image|image|

ref: https://github.com/ruby/typeprof/pull/183

### DIFF
--- a/lib/typeprof/code_range.rb
+++ b/lib/typeprof/code_range.rb
@@ -55,11 +55,11 @@ module TypeProf
       raise unless first
     end
 
-    def self.from_node(node)
+    def self.from_node(node, encoding = Encoding::UTF_16LE)
       node = node.location if node.is_a?(Prism::Node)
       if node.is_a?(Prism::Location)
-        pos1 = CodePosition.new(node.start_line, node.start_column)
-        pos2 = CodePosition.new(node.end_line, node.end_column)
+        pos1 = CodePosition.new(node.start_line, node.start_code_units_column(encoding))
+        pos2 = CodePosition.new(node.end_line, node.end_code_units_column(encoding))
       elsif node.respond_to?(:location)
         loc = node.location
         row, col = loc.start_loc
--- a/scenario/difinitions/definition.rb
+++ b/scenario/difinitions/definition.rb
@@ -1,0 +1,20 @@
+## update: test.rb
+module Animal
+  class Cat
+    def call
+      puts "meow"
+    end
+  end
+end
+
+Animal::Cat.new.call
+
+## definition: test.rb:9:1
+test.rb:(1,7)-(1,13)
+test.rb:(2,8)-(2,11)
+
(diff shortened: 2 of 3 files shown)


lib/typeprof/code_range.rb
```diff
@@ -55,11 +55,11 @@ def initialize(first, last)
       raise unless first
     end
 
-    def self.from_node(node)
+    def self.from_node(node, encoding = Encoding::UTF_16LE)
```
NotFounds, line 58: Is it better to make the encoding configurable during server initialization?
I don’t have a good idea on where to store this option.

scenario/difinitions/definition.rb
```diff
@@ -0,0 +1,20 @@
...
+      puts "meow"
+    end
+  end
+end
+
+Animal::Cat.new.call
+
+## definition: test.rb:9:1
+test.rb:(1,7)-(1,13)
+test.rb:(2,8)-(2,11)
```
NotFounds, line 14: Currently, the definition returns two results on the master branch. I believe that only the first one is accurate. Is this correct?

lib/typeprof/code_range.rb
mame: I think the reason of UTF-16 is because LSP uses it by default, right? I don't think there is any particular need to be able to make it configurable.

scenario/difinitions/definition.rb
mame: This is maybe because it detects not only `Animal` but also `Animal::Cat`. I should fix it, but it's a different problem.

mame:
Thanks! Merged.

Just FYI: I guess there are other issues with the handling of multibyte characters (especially in cases where they have UTF-16 surrogate pairs), perhaps even in handling edit diffs.

https://github.com/ruby/typeprof/blob/b8fd5e196c581dfaf79b29f94582a1316e81264e/lib/typeprof/lsp/text.rb#L33-L34

Since LSP 3.17, we can configure the columns can be represented in UTF-8 code units. In future, I would like to fix this issue by using the LSP negiciation instead of directly handling them in UTF-16.

NotFounds:
Thank you for the review!

> Since LSP 3.17, we can configure the columns can be represented in UTF-8 code units. In future, I would like to fix this issue by using the LSP negiciation instead of directly handling them in UTF-16.

That sounds great!!!

PR #225, opened 2024-06-25
Vertex#show_name is no longer useful for debugging

alpaca-tc:
Removed `Vertex#show_name` because it is no longer useful for debugging.
related: #221 (comment)

I agree with removing it, as I never use it and had more trouble naming `show_name`.


mame:
Awesome, thank you so much!

PR #227, opened 2024-07-10
Support rescue_modifier_node

alpaca-tc:
Implement RescueModifierNode and add test.


mame (approved):

mame:
Sorry for the late action, thank you for the great PR!

PR #228, opened 2024-07-10
Fixed an error that occurred with a call with no method name.

alpaca-tc:
Fixed an error that occurred in the following code.

```ruby
b = -> {}

 # https://github.com/ruby/typeprof//blob/049c296260fa9de3e0298757547c12afd805a693/lib/typeprof/code_range.rb#L71-L71
b.() #=> unknown type: NilClass
```


mame (approved):

mame:
👏 🙏

PR #243, opened 2024-07-23
Support splat (but still very ad-hoc; it only supports a Range)

mame:


mame:
I've been thinking about how to handle splats for method arguments for weeks, but I haven't got it done. I'm going to put it off for now and just do array literal.

PR #259, opened 2024-08-09
Prevent the analysis from getting stuck

mame:
Creating a new vertex in argument matching may cause infinite loop. See scenario/misc/ivar-stuck-case.rb.


mame:
However, this changeset brings a huge performance degeneration. I want a better solution...

PR #272, opened 2024-09-29
Support flip flop operator

long-long-float:
Support flip flop operator.

From [flip flop document](https://docs.ruby-lang.org/ja/latest/doc/spec=2foperator.html#range_cond), I understand as following. If they are incorrect, please let me fix the code.
* The type of flip flop operator is True or False.
* This is only available in conditions.


mame:
I completely missed this PR! Sorry, I'll merge it soon. Thank you very much!

mame (approved):

PR #275, opened 2024-10-10
Fix up gemspec.files

riseshia:
This patch makes gem include files below:

- LICENSE
- README.md
- bin/typeprof
- lib/*
- typeprof.gemspec


mame (approved):

riseshia:
🙏

PR #282, opened 2024-12-16
Support blank ensure statements.

alpaca-tc:
Fixes a `NoMethodError` that occurred when a begin block included a blank ensure clause. Previously, attempting to handle such a statement resulted in the following error:

```
NoMethodError: undefined method `type' for nil
.../typeprof/lib/typeprof/core/ast.rb:24:in `create_node'
.../typeprof/lib/typeprof/core/ast/control.rb:369:in `initialize'
```

Related: https://github.com/test-unit/test-unit/blob/68c6c3a/lib/test/unit/testcase.rb#L623

### DIFF
--- a/lib/typeprof/core/ast/control.rb
+++ b/lib/typeprof/core/ast/control.rb
@@ -366,8 +366,8 @@ module TypeProf::Core
           end
           raw_res = raw_res.subsequent
         end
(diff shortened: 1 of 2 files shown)


mame:
Thank you!

PR #283, opened 2024-12-17
fix: Resolve issue preventing CLI startup with `typeprof --lsp --stdio` option

alpaca-tc:
This commit fixes a bug where the CLI would fail to start when invoked with the `typeprof --lsp --stdio` option.


mame (approved):

mame:
Thank you! 🫣

PR #284, opened 2024-12-18
Support rest arguments

alpaca-tc:
- Initialized rest arguments in method definitions as array.
- Updated method call to add edges to array types.

Further improvements may be possible.


lib/typeprof/core/graph/box.rb
```diff
@@ -519,9 +523,12 @@ def pass_arguments(changes, genv, a_args)
 
         if start_rest < end_rest
           if @f_args.rest_positionals
-            f_arg = @f_args.rest_positionals
             (start_rest..end_rest-1).each do |i|
-              changes.add_edge(genv, a_args.positionals[i], f_arg)
+              @f_args.rest_positionals.each_type do |ty|
+                if ty.is_a?(Type::Instance) && ty.mod == genv.mod_ary && ty.args[0]
```
alpaca-tc, line 528: I noticed some duplicate code, which feels a bit dirty, but I decided to go ahead and open the PR since cleaning it up won't matter if the approach itself is flawed. 
Do you have any suggestions for making this part of the code more DRY? The simplest solution would be to define a private method to consolidate the logic, but it doesn't feel like the optimal approach...

mame:
Thank you! It looks good to me. TBH, I am not sure if this design has any flaw or not. We can reconsider it if we actually use it and find any problem.

As for DRY, I have no strong opinion, but I think this PR is acceptable. The TypeProf source code is very difficult to understand, I guess it may be easier to understand if it is written plainly, even if it is a bit duplicated.

PR #285, opened 2024-12-19
Avoid infinte loop 3

alpaca-tc:
- fix typo `s/infinte/infinite/g`
- Add known-issue scenario for infinite loop
  - Added to known-issues an infinite loop that occurs when parsing the test-unit gem.  It's hard to fix for me 😭 .
  - related: https://github.com/test-unit/test-unit//blob/ef36dc66348753391523be333163c8d5902036b3/lib/test/unit/fixture.rb#L21


mame:
Thanks, that's interesting.

I can see what is happening, but I have no idea how to fix it. I'd like to merge it as a known-issue and take some time to think about it as soon as possible.

alpaca-tc:
I am continuing investigation.

- 1. `B.include A` adds `A` constants to `B`. 
- 2. `:static_read_changed` resolves `A` constant. [here](https://github.com/ruby/typeprof/blob/74c0b3ba61a6d726a743210e2c6a65c947233308/lib/typeprof/core/env/static_read.rb#L48)
  - In this case, it tries to resolve `A` from the scope of `B`.
  - The expected value is `Object::A`.
  - However, due to the relationship of looking for the constant `A` in the order of superclass containing the included modules, `A::A` is found before `Object::A`](https://github.com/ruby/typeprof//blob/74c0b3ba61a6d726a743210e2c6a65c947233308/lib/typeprof/core/env/static_read.rb#L91)
    - This seems to be the cause of the bug.

PR #288, opened 2025-01-01
Move Prism dependency to gemspec file from Gemfile

gemmaro:
Using TypeProf as a library does not resolve `Gemfile` dependencies.  In particular, TypeProf uses the `start_code_units_column` method introduced in [Prism version 0.23.0](https://github.com/ruby/prism/releases/tag/v0.23.0), so it must be at least that version or later.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,8 +7,6 @@ source "https://rubygems.org"
 #  gemspec
 #end
 
-gem "prism", ">= 1.0.0"
-
 group :development do
   gem "rake"
   gem "stackprof", platforms: :mri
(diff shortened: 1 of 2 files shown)


mame:
I'm so sorry for the extremely late reply!

I don't recall the reason why I avoided loading the `gemspec` in the `Gemfile`. I think it's probably fine to load it.
To be honest, I'm not very familiar with `Gemfile` conventions and have mostly been managing it through trial and error.

I did see this PR, but I was hesitant and let it sit because I wanted to avoid adding new dependencies if possible. However, we are already actually depending on it, there's no point in holding out anymore.

Therefore, I agree with the following changes:

* Add `spec.add_runtime_dependency "prism", ">= 1.4.0"` to `typeprof.gemspec`
* Add `gemspec` to the `Gemfile`

I feel terrible for letting this sit for so long, but would you be willing to create a new PR with these changes?
Alternatively, I'm planning to do a release around tomorrow evening, so I can also take care of it myself.

PR #298, opened 2025-04-13
Improve BeginNode Type Handling

smasato:
This PR enhances the BeginNode implementation to properly handle type information in rescue clauses, ensuring accurate type tracking across different execution paths in begin-rescue-else-ensure blocks.

### DIFF
--- a/lib/typeprof/core/ast/control.rb
+++ b/lib/typeprof/core/ast/control.rb
@@ -439,10 +439,72 @@ module TypeProf::Core
 
       def install0(genv)
         ret = Vertex.new(self)
+
+        vars = []
+        @body.modified_vars(@lenv.locals.keys, vars) if @body
+        vars.uniq!
+
+        old_vtxs = {}
+        vars.each do |var|
+          vtx = @lenv.get_var(var)
+          old_vtxs[var] = vtx
+        end
+
         @changes.add_edge(genv, @body.install(genv), ret)
-        @rescue_clauses.each { |clause| @changes.add_edge(genv, clause.install(genv), ret) }
-        @changes.add_edge(genv, @else_clause.install(genv), ret) if @else_clause
+
+        body_vtxs = {}
+        vars.each do |var|
+          body_vtxs[var] = @lenv.get_var(var)
+        end
+
+        clause_vtxs_list = []
+        @rescue_clauses.each do |clause|
+          vars.each do |var|
+            old_vtx = old_vtxs[var]
+            nvtx = old_vtx.new_vertex(genv, self)
+
+            @changes.add_edge(genv, body_vtxs[var], nvtx) unless body_vtxs[var] == old_vtxs[var]
+
+            @lenv.set_var(var, nvtx)
+          end
+
+          @changes.add_edge(genv, clause.install(genv), ret)
+
+          clause_vtxs_list << {}
+          vars.each do |var|
+            clause_vtxs_list.last[var] = @lenv.get_var(var)
+          end
+        end
+
+        if @else_clause
+          vars.each do |var|
+            @lenv.set_var(var, body_vtxs[var])
+          end
+          @changes.add_edge(genv, @else_clause.install(genv), ret)
+          clause_vtxs_list << {}
+          vars.each do |var|
+            clause_vtxs_list.last[var] = @lenv.get_var(var)
+          end
+        end
+
         @ensure_clause.install(genv) if @ensure_clause
+
+        result_vtxs = {}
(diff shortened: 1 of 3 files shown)


mame:
Thank you! I'll take a closer look at it tomorrow, as I don't understand the changes, and it conflicts with #303.

lib/typeprof/core/ast/control.rb
```diff
@@ -392,11 +392,75 @@ def undefine0(genv)
...
       def install0(genv)
         ret = Vertex.new(self)
+
+        vars = []
+        @body.modified_vars(@lenv.locals.keys, vars) if @body
+        @rescue_clauses.each do |clause|
+          clause.modified_vars(@lenv.locals.keys, vars)
+        end
+        @else_clause.modified_vars(@lenv.locals.keys, vars) if @else_clause
+        @ensure_clause.modified_vars(@lenv.locals.keys, vars) if @ensure_clause
```
mame, line 402:
```suggestion
```

Are these lines necessary? I tried deleting them, but the test results did not seem to change.
If you have a test case where these would be needed, it would be helpful if you could indicate it.

lib/typeprof/core/ast/control.rb
```diff
@@ -392,11 +392,75 @@ def undefine0(genv)
...
+
+        @rescue_conds.each { |cond| cond.install(genv) }
+
+        rescue_vtxs = []
+        @rescue_clauses.each do |clause|
+          vars.each do |var|
+            old_vtx = old_vtxs[var]
+            nvtx = old_vtx.new_vertex(genv, self)
+
+            @changes.add_edge(genv, body_vtxs[var], nvtx) if body_vtxs[var]
```
mame, line 426:
```suggestion
            @changes.add_edge(genv, body_vtxs[var], nvtx) unless body_vtxs[var] == old_vtxs[var]
```

I couldn't figure out the case where `body_vtxs[var]` would be falsy.
Also, I think this guard is not needed if lines 398..402 were deleted.

lib/typeprof/core/ast/control.rb
```diff
@@ -392,11 +392,75 @@ def undefine0(genv)
...
+            rescue_vtxs.last[var] = @lenv.get_var(var)
+          end
+        end
+
+        if @else_clause
+          vars.each do |var|
+            @lenv.set_var(var, body_vtxs[var])
+          end
+          @changes.add_edge(genv, @else_clause.install(genv), ret)
+        end
```
mame, line 444:
```suggestion

          rescue_vtxs << {}
          vars.each do |var|
            rescue_vtxs.last[var] = @lenv.get_var(var)
          end
        end
```

It is great that the variable changes in the rescue clause are reflected after the begin node.
Why not reflect the changes in the else clause as well?

Also, it may be good to reflect the changes in the ensure clause.

mame (commented):
Thank you! This PR looks good to me.

I just want to confirm one thing. I wasn't sure about the need for Lines 398..402. Do we need to trace the `modified_vars` in `@rescue_clauses`? If so, it would be nice to add a test scenario.

I feel like I would like to have scenarios to test the other added parts. For example, it would be nice to have a dedicated test scenario for the else clause, since it does a fine job to execute it in the state after body evaluation.

However, if you want, I am OK to add other tests after the merge.

Just FYI, I tested this PR with the following test scenario.

```ruby
## update
def foo
  x = :a
  begin
    x = :b
    raise
    x = :c
  rescue
    check_rescue(x)
    x = :d
  #ensure # Looks like commenting out these two lines makes the test fail? Need to fix
  #  x = :f
  else
    check_else(x)
    x = :e
  end
  check_after(x)
end

def check_rescue(n) = nil
def check_else(n) = nil
def check_after(n) = nil

## assert
class Object
  def foo: -> nil
  def check_rescue: (:a | :c) -> nil
  def check_else: (:c) -> nil
  def check_after: (:a | :c | :d | :e) -> nil
end
```

lib/typeprof/core/ast/control.rb
smasato: It was a lack of understanding. It was unnecessary and I deleted it.

lib/typeprof/core/ast/control.rb
smasato: 398...402 were deleted and are no longer needed.

smasato:
@mame
Thank you for your detailed review!
I have resolved the conflicts and updated and added the test scenario.
Please review when you have time.

mame (approved):

mame:
Thank you!

PR #300, opened 2025-04-15
Support `(class << self; self; end)` syntax used in Minitest

alpaca-tc:
`define_copy` had an error because `@mod_cdef` was not present. Rewrote to re-register module_def process.

```
typeprof/lib/typeprof/core/ast/module.rb:44:in 'TypeProf::Core::AST::ModuleBaseNode#define_copy': undefined method 'add_def' for nil (NoMethodError)

          @mod_cdef.add_def(self)
                   ^^^^^^^^
from typeprof/lib/typeprof/core/ast/base.rb:68:in 'block in TypeProf::Core::AST::Node#define_copy'
```


mame:
Thank you!

PR #302, opened 2025-04-15
Support for implicit rest arguments in array patterns for repl_type_completor

alpaca-tc:
Fix the following error:

```
Error: test: scenario/patterns/array_pat.rb(ScenarioCompiler::ScenarioTest): NoMethodError: undefined method 'expression' for an instance of Prism::ImplicitRestNode

lib/typeprof/core/ast/pattern.rb:8:in 'TypeProf::Core::AST::ArrayPatternNode#initialize'
```


mame:
Thank you!

PR #303, opened 2025-04-17
Support the assignment syntax (=> e) in rescue clauses

coord-e:
TypeProf 264ab6f12528a19403f152def9786f7ec7a804b3 erroneously reports an undefined method to the following code:

```ruby
# rescue-assign.rb
def foo(n)
  raise if n != 0
  n.to_s
rescue StandardError => e
  e.message
end

foo(1)
```

```
$ bin/typeprof rescue-assign.rb --show-errors
# TypeProf 0.30.1

# rescue-assign.rb
# (6,4)-(6,11):undefined method: nil#message
class Object
  def foo: (Integer) -> String?
end
```

`e` in l5 should be recognized as `StandardError`, not `nil`. This PR implements the TypeProf support for the assignment syntax `=> e` in rescue clauses, recognizing proper types for the assigned target `e`.

## ~WIP part~

I found a root cause of my concern, and it should be addressed by a separate PR #304.

<details>

When there's no exception specified in the clause:

```ruby
# rescue-assign-no-excs.rb
def foo(n)
  raise
rescue => e
  e
end

foo(1)
```

I expect `foo` to be `def foo: (Integer) -> untyped` because there should be no edge attached to `@reference.rhs.ret`. However, I get `def foo: (Integer) -> nil` in the current implementation, and I'm not sure how this can happen...

```console
$ bin/typeprof rescue-assign-no-excs.rb --show-errors
# TypeProf 0.30.1

# rescue-assign-no-excs.rb
class Object
  def foo: (Integer) -> nil
end
```

</details>


mame (approved):
Awesome! I said at RubyKaigi a (wrong) solution to break a vertex by using "each_type", but yes, `InstanceTypeBox` is a right solution to fix this issue.

PR #304, opened 2025-04-18
Don't flow nil when there's no else clause in BeginNode

coord-e:
TypeProf https://github.com/ruby/typeprof/commit/264ab6f12528a19403f152def9786f7ec7a804b3 erroneously reports the type of `foo`'s return as nullable: `def foo: -> Integer?`, even though there is no case where `nil` is returned from `foo`.

```ruby
# raise-bot.rb
def foo
  raise
rescue
  1
end
```

```console
$ bin/typeprof raise-bot.rb --show-errors
# TypeProf 0.30.1

# raise-bot.rb
class Object
  def foo: -> Integer?
end
```

`raise` returns `bot` and `bot | Integer` should be equal to `Integer`. I found that the nodes for `else` and `ensure` are coerced to `DummyNilNode` when they are not found, and then they are always connected to `ret`. I think no vertices from `else` or `ensure` should be connected to `ret` when these clauses are not found.


mame (approved):
Good catch, thank you!

PR #305, opened 2025-04-18
Support `it`

White-Green:
TypeProf currently supports `_1` as a block parameter.
This PR also adds support for `it`.


mame (approved):

mame:
Thank you!

Issue #108, opened 2022-01-22
A feature to ignore errors on a specified line


mame:
Kinda `// @ts-expect-error` feature.

```
# Prevent an error report in the next line
1 + "str" # tp-expected-error
```

rhiroe:
Isn't this already resolved by https://github.com/ruby/typeprof/pull/417?

mame:
Yes, thank you for letting me know!

PR #306, opened 2025-04-19
Add `# typeprof:disable/enable` directive for silencing diagnostics

smasato:
## Overview  
This PR adds RuboCop‑style comment directives that let users silence TypeProf diagnostics.

```ruby
foo(1, 2) # typeprof:disable           # inline – only this line

# typeprof:disable                     # block start
foo(1, 2)
bar(1, 2)
# typeprof:enable                      # block end
```

The feature works in CLI and **LSP modes  
CleanShot 2025-04-19 at 21 30 44
CleanShot 2025-04-19 at 21 29 16

---

## Implementation notes
* **`Diagnostic::DisableDirective::Scanner`** – parses comments and returns muted line ranges.  
* **`Diagnostic::DisableDirective::Filter`** – determines whether a diagnostic should be reported.  
* **`ProgramNode#diagnostics`** applies the filter. I implemented the filtering logic in ProgramNode, but please let me know if there is a better way.

---

## Related issue  
Closes [#108](https://github.com/ruby/typeprof/issues/108)

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -15,7 +15,8 @@ module TypeProf::Core
       cref = CRef::Toplevel
       lenv = LocalEnv.new(path, cref, {}, [])
 
-      ProgramNode.new(raw_scope, lenv)
+      disable_ranges = TypeProf::Diagnostic::DisableDirective::Scanner.collect(result, src)
+      ProgramNode.new(raw_scope, lenv, disable_ranges: disable_ranges)
     end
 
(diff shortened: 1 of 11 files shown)


smasato:
@mame 
Thanks for the comment!

I followed RuboCop’s style, so `# typeprof:disable/enable` only turns diagnostics on or off for that block—it never touches the type‑inference logic itself. That wording might still confuse users, so I’m open to better keywords. Let me know what you think whenever you’re ready.

PR #309, opened 2025-04-24
Fixed typeprof.conf.json could not be loaded on VSCode on Windows

S-H-GAMELINKS:
The problem was that the `url` processed by `uri_to_path` was like `c%3A/Users/shunh/vscode-typeprof-test`, and typeprof.conf.json could not be loaded.

This meant that TypeProf could be launched as an LSP, but the results of type analysis were not displayed on VSCode.

So, I changed that if `%3A` is included in the url returned by `uri_to_path`, it is replaced with `:`.
However, if you have any other good ways to fix it, please comment.

Note:
- Environment
  - OS: Windows 11
  - Ruby: Ruby 3.4.3(installed by RubyInstaller)
  - VSCode 1.99.3

### DIFF
--- a/lib/typeprof/lsp/server.rb
+++ b/lib/typeprof/lsp/server.rb
@@ -69,7 +69,13 @@ module TypeProf::LSP
     end
 
     def uri_to_path(url)
-      url.delete_prefix(@url_schema)
+      path = url.delete_prefix(@url_schema)
+
+      if path.include?('%3A')
+        path.sub(/%3A/, ':')
+      else
+        path
+      end
     end
 
     #: (Array[String]) -> void
--- a/lib/typeprof/lsp/server.rb
+++ b/lib/typeprof/lsp/server.rb
@@ -1,3 +1,5 @@
+require "cgi"
+
 module TypeProf::LSP
   module ErrorCodes
     ParseError = -32700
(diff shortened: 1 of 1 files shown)


mame:
Hmm, should we use `CGI.unescape_uri_component`?

S-H-GAMELINKS:
Thank you for your comment.
I tried `CGI.unescape_uri_component` and it worked fine.
So I replaced it with that.

mame:
Ah, maybe we should add `CGI.escape_uri_component` in path_to_uri as well.
Can you check if the following implementation works on your Windows environment?

```ruby
def path_to_uri(path)
  @url_schema + File.expand_path(path).split("/").map {|s| CGI.escape_uri_component(s) }.join("/")
end

def uri_to_path(url)
  uri.delete_prefix(@url_schema).split("/").map {|s| CGI.unescape_uri_component(s) }.join("/")
end
```

S-H-GAMELINKS:
It worked fine.
I'll commited this code.

smasato:
Maybe the cgi gem bundled with Ruby 3.2 is the version before the `CGI.escape_uri_component` alias was implemented.
https://github.com/ruby/cgi/pull/27
For compatibility you need to use `CGI.escapeURIComponent` / `CGI.unescapeURIComponent`.

Aside from that, TypeProf does not support Ruby 3.2 and can be excluded from the CI matrix.

S-H-GAMELINKS:
Thank you for your comment.

I thought it would be better to modify to use CGI.escapeURIComponent / CGI.unescapeURIComponent as a fix for Windows.

Also, it would be better to deal with removing Ruby 3.2 from CI as a separate issue, such as by creating a separate PR.

smasato:
> Also, it would be better to deal with removing Ruby 3.2 from CI as a separate issue, such as by creating a separate PR.

I think so too.
I created https://github.com/ruby/typeprof/pull/313 .

mame (approved):

mame:
Thanks!

PR #310, opened 2025-04-25
feat: add --init option

ahogappa:
I added an option to TypeProf to generate a configuration file.
When this option is used, it creates the config file automatically.

However, what I really want is to generate the config file through a VS Code command—like what the [Solargraph](https://github.com/castwide/vscode-solargraph?tab=readme-ov-file#project-configuration) extension does.
It would be odd to put the file generation logic in the VS Code extension itself, so I added it as a TypeProf option instead.
If this proposal is accepted, I’d like to submit a PR to add a command to the extension as well.

### DIFF
--- a/lib/typeprof/cli/cli.rb
+++ b/lib/typeprof/cli/cli.rb
@@ -11,6 +11,7 @@ module TypeProf::CLI
 
       output = nil
       rbs_collection_path = nil
+      initialize_config_file = false
 
       opt.separator ""
       opt.separator "Options:"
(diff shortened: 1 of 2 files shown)


lib/typeprof/cli/cli.rb
```diff
@@ -175,5 +182,9 @@ def set_profiler
         StackProf.results
       end
     end
+
+    def generate_config_file
+      File.write('typeprof.conf.jsonc', File.read(File.join(__dir__, 'typeprof.conf.jsonc')))
```
mame, line 187:
```suggestion
      File.write('typeprof.conf.jsonc', File.read(File.join(__dir__, 'typeprof.conf.jsonc')), mode: "wx")
```

I think it's a good idea to raise an error when trying to overwrite an existing file

mame (commented):
Thank you! The configuration file (typeprof.conf.jsonc) is still tentative, but I think it makes sense to have `typeprof --init` command and a mechanism to call it from the vsode extension!

I think we should prohibit overwriting, what do you think?

It would be better if we have a test, but not required.

lib/typeprof/cli/cli.rb
ahogappa: Thank you! This one is definitely much better.

PR #312, opened 2025-04-27
Update Prism to 1.4.0 and remove workarounds

smasato:
This PR updates the Prism dependency from 1.0.0 to 1.4.0 and removes workarounds that were needed for the old version.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ source "https://rubygems.org"
 #  gemspec
 #end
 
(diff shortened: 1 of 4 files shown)


mame (approved):
Since Ruby 3.4 bundles prism 1.2.0 by default, I was wondering that I should get stuck with prism 1.2.0... But If I asked myself "should I add a CI matrix for prism 1.2.0?", I don't want so. So I will merge it!

PR #314, opened 2025-04-27
Update README.md

smasato:
This PR adds information to README.md and organizes it.

Is there any other information I should add?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,31 @@
 
 An experimental type-level Ruby interpreter for testing and understanding Ruby code.
 
+## Installation
+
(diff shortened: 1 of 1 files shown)


mame:
Great, thank you! Now that a new command `typeprof --init` has been proposed (#310) and I will merge it soon. It would be better to write that procedure instead of manually putting `typeprof.conf.json`.

smasato:
@mame 
I replaced the procedure about configuration files with the use of `typeprof --init`.
I also added a link to the marketplace for the VS Code extension.

mame (approved):

PR #320, opened 2025-04-30
Fix handling of PostExecutionNode

smasato:
This PR fixes PostExecutionNode so that it can be evaluated in the correct order.

Is `nil` the correct return value for the foo method in the test scenario? If so, fix it.


mame:
Thanks!

It might sound strange to say this after I wrote the scenario file, but I’m a bit unsure whether we really need to seriously support END in method, since Ruby shows a warning when it's used.

```
$ ruby -e 'def foo; END {}; end'
-e:1: warning: END in method; use at_exit
```

To be precise, I feel like this implementation might not be perfect. That's because END is not executed at the end of the scope, but at the end of the method.

```ruby
def foo
  x = nil
  1.times do
    END { check(x) } # This passes a String (not an Integer)
    x = 42
  end
  x = "str"
end
```

Also, since END is executed even when exiting via early return or exceptions, if we wanted to handle it accurately, we'd probably need to treat it like an ensure clause. But honestly, I think it is too much.

Since you went through the trouble of writing it, I'll merge it for now, but I might remove this handling in the future. Sorry in advance if that happens!

PR #321, opened 2025-05-01
Add a scenario for anonymous block forwarding

sinsoku:
I noticed that TestProf does not work with code that uses anonymous block forwarding, so I added a scenario.

### DIFF
--- a/scenario/block/anonymous_block_forwarding.rb
+++ b/scenario/block/anonymous_block_forwarding.rb
@@ -1,0 +1,18 @@
+## update
+def foo(&)
+  bar(&)
+end
+
+def bar(&b)
+  b.call(1.0)
+end
+
+foo do |n|
+  "str"
+end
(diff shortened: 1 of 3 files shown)


scenario/block/anonymous_block_forwarding.rb
```diff
@@ -0,0 +1,18 @@
...
+end
+
+foo do |n|
+  "str"
+end
+
+## assert
+class Object
+  def foo: -> untyped
+  def bar: { (Float) -> untyped } -> untyped
```
sinsoku, line 17: I really want to generate the following RBS, but I couldn't fix it.

```rbs 
def foo: { (Float) -> String } -> String 
def bar: { (Float) -> String } -> String 
```

I could at least avoid the NoMethodError by checking `raw_block.expression`, so I made a PR.

PR #322, opened 2025-05-01
Support RBS::AST::Members::InstanceVariable

S-H-GAMELINKS:
When I was implementing TypeProf LSP plugin for coc.nvim, and I found the following this error.

```bash
'TypeProf::Core::AST.create_rbs_member': unsupported: RBS::AST::Members::InstanceVariable (RuntimeError)
```

Reading the code based on the error message, I found that an error occurs when instance variable written in RBS.

Therefore, I have added the `SigInstanceVariableNode` class that can be used in the case of `RBS::AST::Members::InstanceVariable` in `TypeProf::Core::AST.create_rbs_member` method.

However, I am not familiar with the internal code of TypeProf, so I would appreciate any comments if you have any better ways of writing it.

### DIFF
--- a/lib/typeprof/core/ast/sig_decl.rb
+++ b/lib/typeprof/core/ast/sig_decl.rb
@@ -434,6 +434,45 @@ module TypeProf::Core
       end
     end
 
+    class SigInstanceVariableNode < Node
+      def initialize(raw_decl, lenv)
+        super(raw_decl, lenv)
+        @var = raw_decl.name
+        @cpath = lenv.cref.cpath
+        @scope_level = lenv.cref.scope_level
+        @type = AST.create_rbs_type(raw_decl.type, lenv)
+      end
+
+      attr_reader :cpath, :scope_level, :type
+      def subnodes = { type: }
+      def attrs = { cpath: }
+
+      def define0(genv)
+        @type.define(genv)
+        mod = genv.resolve_ivar(cpath, scope_level == :class, @var)
+        mod.add_decl(self)
+        mod
+      end
+
+      def define_copy(genv)
+        mod = genv.resolve_ivar(cpath, scope_level == :class, @var)
+        mod.add_decl(self)
+        mod.remove_decl(@prev_node)
+        super(genv)
+      end
+
+      def undefine0(genv)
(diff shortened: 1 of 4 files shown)


lib/typeprof/core/ast/sig_decl.rb
```diff
@@ -434,6 +434,45 @@ def install0(genv)
...
+        @type = AST.create_rbs_type(raw_decl.type, lenv)
+      end
+
+      attr_reader :cpath, :scope_level, :type
+      def subnodes = { type: }
+      def attrs = { cpath: }
+
+      def define0(genv)
+        @type.define(genv)
+        mod = genv.resolve_ivar(cpath, scope_level == :class, @var)
```
sinsoku, line 452: Since this is an instance variable type declaration, it is always `false` here.
However, `@scope_level` returns `:class`, so it is `true`.

It would be better to always pass false.

```ruby
mod = genv.resolve_ivar(cpath, false, @var)
```

sinsoku (commented):
How about adding a scenario for instance variable type declarations?
Tests for RBS are in scenario/rbs.

lib/typeprof/core/ast/sig_decl.rb
mame: Right, the original PR failed with the following scenario.

```
## update: test.rbs
class Foo
  @foo: String
end

## update: test.rb
class Foo
  def check
    @foo
  end
end

## assert
class Foo
  def check: -> String
end
```

```
Failure: test: t.rb(ScenarioCompiler::ScenarioTest)
t.rb:13:in 'block in <class:ScenarioTest>'
<"class Foo\n" + "  def check: -> String\n" + "end"> expected but was
<"class Foo\n" + "  def check: -> untyped\n" + "end">
```

I imported your suggestion. https://github.com/ruby/typeprof/pull/322/commits/0cbe1492733d3e94cef532d3bb09edf02bdbeafe

mame:
@S-H-GAMELINKS Thank you very much for the contribution and very sorry for my very late response! I will merge it soon

S-H-GAMELINKS:
@mame 
I'm sorry I didn't get back to you too 
Thanks again merge!

PR #329, opened 2025-06-24
Remove eval workaround for keyword arguments in SigAttrReaderNode and SigAttrWriterNode

smasato:
This PR removes the `eval` workaround that was used to prevent TypeProf from failing to parse keyword arguments during dogfooding. Since TypeProf now supports keyword arguments, this workaround is no longer necessary.


mame (approved):
Thanks!

PR #333, opened 2025-08-09
Fix unknown type variable error in generic RBS class definitions

sinsoku:
Resolves the "unknown type variable: X" error that occurred when processing instance variables using type parameters within generic RBS class definitions like `class Foo[X]` with `@x: X`.

The fix creates proper type parameter substitution maps for SigInstanceVariableNode contexts, allowing type variables to be correctly resolved within generic classes.

🤖 Generated with [Claude Code](https://claude.ai/code)


mame:
Sorry for the delayed action. I was quite troubled about this, but I decided to merge it and consider it later :-)

PR #338, opened 2025-09-15
Fix NoMethodError in SigTyIntersectionNode when using intersection types

sinsoku:
When I tried TypeProf in VSCode, I got the following error:

```
/Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:503:in `typecheck': undefined method `each' for nil (NoMethodError)

        @types.each do |type|
              ^^^^^
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/graph/box.rb:152:in `block in match_arguments?'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/graph/box.rb:151:in `each'
```

This is because the SigTyIntersectionNode class does not have an initialize method, and the `@types` instance is not initialized.
This commit fixes it.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/lib/typeprof/core/ast/sig_type.rb
+++ b/lib/typeprof/core/ast/sig_type.rb
@@ -491,12 +491,25 @@ module TypeProf::Core
     end
 
     class SigTyIntersectionNode < SigTyNode
+      def initialize(raw_decl, lenv)
+        super(raw_decl, lenv)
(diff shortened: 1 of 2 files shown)


mame (approved):

mame:
Thanks!

PR #340, opened 2025-09-15
Fix NoMethodError in `TypeProf::Core::Type::Singleton`

sinsoku:
I encountered the following error:

```
/Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:10:in `block in typecheck_for_module': undefined method `args' for an instance of TypeProf::Core::Type::Singleton (NoMethodError)

            f_args.zip(ty.args) do |f_arg_node, a_arg_ty|
                         ^^^^^
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/graph/vertex.rb:11:in `each_key'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/graph/vertex.rb:11:in `each_type'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:5:in `typecheck_for_module'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:669:in `typecheck'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:483:in `block in typecheck'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:482:in `each'
        from /Users/sinsoku/ghq/github.com/ruby/typeprof/lib/typeprof/core/ast/sig_type.rb:482:in `typecheck'
```

The `args` method is only defined on `Type::Instance`, so it adds a conditional expression to avoid the error.

### DIFF
--- a/scenario/regressions/singleton-is-a-module.rb
+++ b/scenario/regressions/singleton-is-a-module.rb
@@ -1,8 +1,2 @@
 ## update
-class C
-  def self.foo=(v)
-    v.is_a?(Module)
-  end
-
-  self.foo = 1
-end
+1.is_a?(Module)


scenario/regressions/singleton-is-a-module.rb
```diff
@@ -0,0 +1,8 @@
+## update
+class C
+  def self.foo=(v)
+    v.is_a?(Module)
+  end
+
+  self.foo = 1
+end
```
mame: I think we can simplify this test case.

```suggestion
## update
1.is_a?(Module)
```

PR #341, opened 2025-09-27
Fix infinite loop caused by inline RBS comments

sinsoku:
When analyzing the following code with TypeProf, an infinite loop occurs.

```ruby
class C
  #: (Array[untyped]) -> Array[untyped]
  def foo(x)
    x
  end
end
```

TypeProf generates duplicate vertices for the same sig node.
To resolve this issue, it restores the code removed in 3251cbc362112381f5e84f51ba6f82c64dc170ac.

---

I'm not sure if restoring this code is correct.
However, all tests now pass, so I created a PR.

### DIFF
--- a/lib/typeprof/core/graph/change_set.rb
+++ b/lib/typeprof/core/graph/change_set.rb
@@ -22,7 +22,7 @@ module TypeProf::Core
       @new_depended_superclasses = []
     end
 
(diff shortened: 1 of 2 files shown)


mame:
Good catch! I thought the hack was no longer needed, but actually needed. I will give another try to remove it later

PR #342, opened 2025-10-03
Output the file path when an error occurs during analysis

sinsoku:
Add error handling to output the path of the file being analyzed when an
exception occurs, making it easier to identify which file caused the error.


mame:
This looks so useful!

PR #346, opened 2025-11-02
Add the `typeprof.enableToggleButton` notification

sinsoku:
The toggle button is not displayed in vscode-typeprof because there is
no `typeprof.enableToggleButton` notification.
refs: https://github.com/ruby/vscode-typeprof/blob/7431f0ea6775e5013fb47bc96b74ad5d4d91584a/src/extension.ts#L287-L289

This notification was in v0.21.11 but is no longer there.
refs: https://github.com/ruby/typeprof/blob/v0.21.11/lib/typeprof/lsp.rb#L355

Add a notification to display the toggle button.


mame:
Thank you!

PR #356, opened 2025-12-25
Fix crash in break in loop

pvcresin:
This PR fixes an error that occurs when breaking out of a loop.

## Reproduction

The following code crashes when analyzed.

```rb
def foo
  count = 0
  loop do
    count += 1
    begin
      break count if count == 3
    rescue
      break count
    end
  end
end
```

## Approach

A bridge Vertex is added between src and dst to prevent duplication.


mame:
Thank you, and sorry for the delayed response! I plan to actively respond starting today.

I don't think your change is bad, but I have some reservations about implicitly adding vertexes when a duplicate occurs. I'd like to handle this on the BreakNode side for now.

PR #363, opened 2026-02-08
Prevent `Hash#[]=` on local variables from polluting parameter types

### LINKED ISSUE #362: Parameter type includes values assigned via `Hash#[]=` on local variable
## Steps to reproduce

Analyze the following file with TypeProf.

```ruby
def foo(options)
  options[:name] = "str"
  nil
end

foo(Hash.new)
```

### Expected behavior

```rbs
class Object
  def foo: (Hash[untyped, untyped]) -> nil
end
```

Parameter types should only reflect the types passed from call sites, not values assigned inside the method.

### Actual behavior

```bash
$ bin/typeprof reproduce.rb
# TypeProf 0.31.1

# reproduce.rb
class Object
  def foo: (Hash[:name, String]) -> nil
end
```

### System configuration

**Ruby version**: ruby 4.0.0 (2025-12-25 revision 553f1675f3) +PRISM [arm64-darwin25]
**TypeProf version**: 753de9b8501f4f7c509e605ea9c9982efa8fc311

## Real-world use cases

I noticed this issue when analyzing Action Pack with TypeProf.

The minimum steps required to actually analyze Action Pack are as follows:

```bash
$ bin/typeprof tmp/rails/actionpack/test/controller/routing_test.rb \
  tmp/rails/actionpack/lib/action_dispatch/routing/route_set.rb \
  tmp/rails/actionpack/lib/action_dispatch/journey/formatter.rb > ap.rbs

$ cat ap.rbs | grep url_for | wc -c
   13936

$ cat ap.rbs | grep url_for
      def url_for: (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], (Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], String] | Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], untyped] | String)?] | Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | :path | :path_params | :person | :port | :protocol | :script_name | :subdomain | :tld_length | :trailing_slash | :use_route | :user | Range[Integer], String] | Hash[:_recall | :action | :anchor | :controller | :domain | :foo | :format | :host | :id | :name | :only_path | :original_script_name | :params | :password | ...(omitted
```

The parameter type contains a recursive structure such as `Hash[:_recall | ..., Hash[:_recall | ..., Hash[:_recall, ...]`.
I have identified the cause of this as `Hash#[]=`.

sinsoku:
When `Hash#[]=` was called on a local variable inside a method, the
assigned key/value types flowed back into the method's parameter type.
This caused type explosion in real-world code like Action Pack's
`url_for`, where the parameter type grew into a deeply nested recursive
Hash type.

Introduce HashAsetBox to handle `Hash#[]=` on local variables with
flow-sensitive tracking. Instead of modifying the original hash type
via backflow, it creates a new variable version with the updated type,
so the parameter type only reflects types from call sites.

fixes #362

### DIFF
--- a/lib/typeprof/core/ast/call.rb
+++ b/lib/typeprof/core/ast/call.rb
@@ -164,6 +164,15 @@ module TypeProf::Core
           @changes.add_edge(genv, allow_nil, ret)
         end
 
+        if @mid == :[]= && @recv.is_a?(LocalVariableReadNode)
+          key_node = @positional_args[0]
+          if key_node.is_a?(SymbolNode)
+            recv_vtx = @lenv.get_var(@recv.var)
+            nvtx = @lenv.new_var(@recv.var, self)
+            @changes.add_hash_aset_box(genv, recv_vtx, key_node.lit, ret, nvtx)
+          end
+        end
+
         ret
       end
 
@@ -188,6 +197,10 @@ module TypeProf::Core
       end
 
       def modified_vars(tbl, vars)
+        if @mid == :[]= && @recv.is_a?(LocalVariableReadNode) && tbl.include?(@recv.var)
+          key_node = @positional_args[0]
+          vars << @recv.var if key_node.is_a?(SymbolNode)
+        end
         subnodes.each do |key, subnode|
           next unless subnode
           if subnode.is_a?(AST::Node)
--- a/lib/typeprof/core/builtin.rb
+++ b/lib/typeprof/core/builtin.rb
@@ -111,9 +111,16 @@ module TypeProf::Core
 
     def hash_aset(changes, node, ty, a_args, ret)
       if a_args.positionals.size == 2
+        val = a_args.positionals[1]
+
+        # Skip backflow for local variable receivers (handled by HashAsetBox)
+        if node.recv.is_a?(AST::LocalVariableReadNode)
+          changes.add_edge(@genv, val, ret)
+          return true
+        end
+
         case ty
         when Type::Hash
-          val = a_args.positionals[1]
           idx = node.positional_args[0]
           if idx.is_a?(AST::SymbolNode) && ty.get_value(idx.lit)
             # TODO: how to handle new key?
--- a/lib/typeprof/core/graph/box.rb
+++ b/lib/typeprof/core/graph/box.rb
@@ -1136,25 +1136,16 @@ module TypeProf::Core
     end
 
     def run0(genv, changes)
+      has_record = false
+
       @recv.each_type do |ty|
         case ty
         when Type::Record
-          new_fields = {}
+          has_record = true
           ty.fields.each do |key, field_vtx|
             @field_cache[key] ||= Vertex.new(@node)
             changes.add_edge(genv, field_vtx, @field_cache[key]) unless field_vtx.equal?(@field_cache[key])
-            new_fields[key] = @field_cache[key]
           end
-          @field_cache[@key_sym] ||= Vertex.new(@node)
-          new_fields[@key_sym] = @field_cache[@key_sym]
-          changes.add_edge(genv, @val_vtx, @field_cache[@key_sym])
-          new_fields.each do |key, vtx|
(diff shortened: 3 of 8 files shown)


sinsoku:
@mame 
Sorry, the tests never finished. 😢 
The changes in this PR may have put `tool/dog_bench.rb` into an infinite loop.
I was unable to cancel the job.

The default timeout for GitHub Actions is 360 minutes, so it would be helpful if you could cancel it before it wastes execution time.

mame:
Sorry for leaving this for six months!

I found two problems.

```ruby
h = Hash.new
h["s"] = 42
# master:  Hash[String, Integer]
# this PR: Hash[untyped, untyped]
```

`hash_aset`'s early return only checks the receiver; `call.rb` also requires a symbol literal key.

```ruby
def foo(opts, m)
  while m
    opts[:k0] = opts
    opts[:k1] = opts
    opts[:k2] = opts
    opts[:k3] = opts
  end
  opts
end
foo({}, 1)
# master:  52B
# this PR: 45,578,097B, dump 5.2s   (five keys: never returns)
```

I tried dropping `@field_cache` and reusing the receiver's own field vertices — 45MB down to 10KB. But then:

```ruby
def foo(flag)
  h = {}
  h[:a] = 1 if flag
  # ... 15 of them
  h
end
#          this PR  |  with my change
#   n=5      253B   |    1,163B
#   n=10     758B   |   66,603B
#   n=15   1,563B   | 3,113,003B
```

So `@field_cache` is what keeps that in check. I can't think of a fix that avoids both.

PR #367, opened 2026-02-09
Fix infinite loop when block has multiple params for sort_by etc

pvcresin:
This PR fixes an infinite loop that occurs when a multi-parameter block is passed to methods like `sort_by`.

## Reproduction

The following code causes the analysis to hang.

```rb
def OK_1
  { a: 1 }.sort_by do |key_value|
    key_value
  end
end

def OK_2
  { a: 1 }.sort_by do |(key, value)|
    key
  end
end

def NG
  { a: 1 }.sort_by do |key, value|
    key
  end
end
```

## Approach

SplatBox creation is moved from the run phase (`Block#accept_args`) to the install phase (`CallBaseNode#install0`), and `accept_args` now just connects an edge to the pre-existing `@f_ary_arg` vertex. This keeps the graph structure stable across re-runs and prevents the infinite loop caused by typecheck trigger edges.

### DIFF
--- a/lib/typeprof/core/ast/call.rb
+++ b/lib/typeprof/core/ast/call.rb
@@ -139,7 +139,11 @@ module TypeProf::Core
           end
 
           blk_f_ary_arg = Vertex.new(self)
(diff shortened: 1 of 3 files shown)


mame (approved):
👏

PR #369, opened 2026-02-16
Fix hover on method parameters with inline RBS annotations

whtsht:
There was an issue where hovering was not possible when method arguments had rbs comments.

The root cause was an off-by-one error: `Prism::Location#start_line` is 1-based, but `rbs_comments` is indexed by 0-based line numbers.
As a result, the first element of `rbs_comments` ended up being `nil`, e.g.:

```
[nil, " (Integer) -> void"]
```
https://github.com/whtsht/typeprof/commit/0e6baab855ad21e7b04cf9cb7eb7709c630d102f

This PR fixes the issue by subtracting 1 from `start_line` before indexing/storing into `rbs_comments`.


mame (approved):
Thank you very much!

PR #370, opened 2026-02-17
Fix fcall include handling with context-aware meta parsing

pvcresin:
This PR fixes `scenario/known-issues/fcall-include.rb`.

The known issue was this kind of code:

```ruby
class C
  def include(x)
    x
  end

  def check
    include(1)
  end
end
```

TypeProf incorrectly treated `include(1)` as meta `include` in this context, so type inference became too loose.

## Details

This PR fixes it by changing AST parsing:

- Receiver-less `include` / `attr_reader` / `attr_accessor` are treated as meta only at class/module top level.
- Calls in method bodies are parsed as normal method calls.

After this change:

- The old known issue is moved to `scenario/regressions/fcall-include.rb`.
- The scenario now passes as a regression test.


mame (approved):
Neat solution!

PR #372, opened 2026-02-17
Fix keyword-as-hash handling for methods without keyword params

pvcresin:
This PR fixes TypeProf’s handling of calls like `foo(x: 1)` when the callee does not declare keyword parameters.(`scenario/known-issues/keywords.rb`.)  

```ruby
def foo(a)
  a
end

foo(x: 1)
```

In those cases, keyword arguments are now treated as a trailing positional hash (Ruby-compatible), improving inference from `untyped` to record/hash types.

## Details

- Track `Prism::NoKeywordsParameterNode` (`**nil`) as `no_keywords` in method AST metadata.
- Add `ActualArguments#with_keywords_as_last_positional_hash` to centralize argument normalization.
- Normalize keyword args in `MethodDefBox#pass_arguments` only when:
  - keyword args are present,
  - the method has no required/optional/rest keyword params,
  - and the method is not `**nil`.
- Promote the scenario from `scenario/known-issues/keywords.rb` to `scenario/args/keyword-as-hash.rb`.

### DIFF
--- a/lib/typeprof/core/ast/method.rb
+++ b/lib/typeprof/core/ast/method.rb
@@ -37,6 +37,7 @@ module TypeProf::Core
           opt_keywords: [],
           opt_keyword_defaults: [],
           rest_keywords: nil,
+          no_keywords: false,
           block: nil,
         }
       end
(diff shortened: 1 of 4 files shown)


mame (approved):
:clap::clap::clap:

PR #430, opened 2026-04-10
Fix NoMethodError in SigTySingletonNode#show due to uninitialized @args

sinsoku:
SigTySingletonNode#initialize did not initialize @args, causing
a NoMethodError when #show was called during return type checking.
This occurred when a method annotated with singleton(X) returned
a mismatched type.

### DIFF
--- a/lib/typeprof/core/ast/sig_type.rb
+++ b/lib/typeprof/core/ast/sig_type.rb
@@ -550,7 +550,8 @@ module TypeProf::Core
         name = raw_decl.name
         @cpath = name.namespace.path + [name.name]
         @toplevel = name.namespace.absolute?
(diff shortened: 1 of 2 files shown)


mame:
Thank you. It seems that `RBS::Types::ClassSingleton#args` was added in RBS 4.0. I wonder whether we should require rbs >= 4.0.0, or branch using `respond_to?` instead.

sinsoku:
Thank you for the review.

Since RBS 4.0 was recently released, I think we should maintain compatibility with RBS 3.x for now. I've updated the code to use `respond_to?(:args)` to avoid errors on RBS 3.x.

mame (approved):
Thanks!

PR #433, opened 2026-04-13
Lazily initialize ChangeSet collections to reduce GC pressure

sinsoku:
# Summary

`ChangeSet#initialize` was allocating 7 empty Hashes on every
instantiation, accounting for 13.2% of CPU samples. Most of these
collections are never used in a given ChangeSet instance.

Remove the eager Hash allocations from initialize and use lazy
accessor methods (e.g. `def edges = @edges ||= {}`) to allocate on
first use. Array fields (`@diagnostics`, `@depended_*`) and the
if-target guard are kept as-is to preserve the existing contract.

## Benchmark

Benchmark on Redmine v6.1.1 (app + sig):
- Before: ~19.0s, 62 GCs, 34.1M allocations
- After:  ~14.9s, 47 GCs, 31.7M allocations (~22% faster)

Measured with [bench_redmine.rb](https://gist.github.com/sinsoku/bff00848fdd341512ac3edbd9f07c69a).

### DIFF
--- a/lib/typeprof/core/graph/change_set.rb
+++ b/lib/typeprof/core/graph/change_set.rb
@@ -3,13 +3,6 @@ module TypeProf::Core
     def initialize(node, target)
       @node = node
       @target = target
-      @new_vertexes = {}
-      @covariant_types = {}
-      @contravariant_types = {}
-      @edges = {}
-      @new_edges = {}
-      @boxes = {}
-      @new_boxes = {}
       @diagnostics = []
       @new_diagnostics = []
       if target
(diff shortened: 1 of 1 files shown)


mame:
I think lazily initializing `@new_vertexes`, `@covariant_types`, and `@contravariant_types` makes sense since they can indeed be unnecessary. However, `@edges`, `@new_edges`, `@boxes`, and `@new_boxes` seem to always get initialized in `reinstall`. I wonder if there are cases where a `ChangeSet` is created but `reinstall` is never called...?

sinsoku:
To be honest, I simply delayed initializing the instance variables in `initialize` without thinking deeply about it.
After your review, I looked into it again and found that when `install_copy` / `copy_from` is used instead of `install`, there are cases where `reinstall` is not called.

mame (approved):
I give it a try!

PR #435, opened 2026-04-13
Lazily initialize BasicVertex and MethodEntity to reduce GC pressure

sinsoku:
`BasicVertex#initialize` and `MethodEntity#initialize` were eagerly
allocating collections that are often unused, contributing to GC
pressure (38.1% of CPU samples in `tool/dog_bench.rb`).

- BasicVertex: remove eager `@types_to_be_added = {}` and use `||=`
  on first write in `on_type_added`
- MethodEntity: replace 4 eager Sets and 1 Hash with lazy accessor
  methods

Measured with `tool/dog_bench.rb` (stackprof):
- GC overhead: 38.1% -> 23.7%
- Total samples: 1132 -> 1001

### DIFF
--- a/lib/typeprof/core/env/method_entity.rb
+++ b/lib/typeprof/core/env/method_entity.rb
@@ -2,60 +2,63 @@ module TypeProf::Core
   class MethodEntity
     def initialize
       @builtin = nil
-      @decls = Set.empty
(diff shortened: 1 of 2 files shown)


mame (approved):
Let's give it a try!

PR #437, opened 2026-04-14
Fix infinite loop in `define_all` for RBS include with same-name nested constant

### LINKED ISSUE #431: http-2 signature causes infinite loop
When typeprof loads an "http-2" signature via the "gem_rbs_collection", an infinite loop occurs and initialization fails to complete.

## Reproducing

```ruby
# Gemfile
source "https://rubygems.org"

gem "typeprof", github: "ruby/typeprof"
gem "http-2"
```

```ruby
# foo.rb
class Foo
  def foo
    "foo"
  end
end
```

bundler setup and execute:

`rbs collection init && rbs collection install && typeprof foo.rb` 

typeprof hangs up and seems to be in an infinite loop

sinsoku:
When `module X` contains `class X`, resolving `include X` in RBS
searched the ancestor chain (`search_ancestors=true`), which caused the
resolved constant to flip between X (module) and X::X (class) on
every re-evaluation, looping `define_all` forever.

The .rb side already avoids this via `use_strict_const_scope` in
IncludeMetaNode. Apply the same `strict_const_scope=true` to
SigIncludeNode and SigPrependNode so that include/prepend targets
are resolved through lexical scope only.

fixes #431

### DIFF
--- a/lib/typeprof/core/ast/sig_decl.rb
+++ b/lib/typeprof/core/ast/sig_decl.rb
@@ -232,10 +232,10 @@ module TypeProf::Core
       def define0(genv)
         @args.each {|arg| arg.define(genv) }
         const_reads = []
-        const_read = BaseConstRead.new(genv, @cpath.first, @toplevel ? CRef::Toplevel : @lenv.cref, false)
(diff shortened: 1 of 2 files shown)


mame (approved):
Thanks!

PR #439, opened 2026-04-22
Add report guide for writing scenario files

sinsoku:
Follow-up to #376.

Instead of formal issue templates, this adds a lightweight guide (doc/report\_guide.md) to make it easy to report issues. It covers how to write and run scenario files, which users can optionally reference when reporting bugs or proposing features.


mame (approved):
Great!

PR #440, opened 2026-04-23
Add known issues scenarios

pvcresin:
I have added a scenario that includes code that doesn't seem to be supported.

Additions, modifications, and deletions are welcome.


mame (approved):
Need a lot of work!

PR #441, opened 2026-04-24
Fix crash on empty parentheses () in AST conversion

ahogappa:
## Problem

  `TypeProf::Core::AST.create_node` crashes with `NoMethodError` on the empty parentheses expression `()`:

  ```ruby
  require "typeprof"
  TypeProf::Core::Service.new({}).update_rb_file("t.rb", "()")
  # => lib/typeprof/core/ast.rb:62:in 'create_node':
  #    undefined method 'type' for nil (NoMethodError)
  #            case raw_node.type
```

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -61,6 +61,7 @@ module TypeProf::Core
       while true
         case raw_node.type
         when :parentheses_node
+          return DummyNilNode.new(lenv.code_range_from_node(raw_node), lenv) if raw_node.body.nil?
           raw_node = raw_node.body
         when :implicit_node
           raw_node = raw_node.value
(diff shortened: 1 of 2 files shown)


sinsoku:
I think it's a good change.

It might also be worth testing a couple more AST paths where `DummyNilNode` is received by a different parent node:

- `[()]` — `DummyNilNode` as an `ArrayNode` element
- `if (); 1; end` — `DummyNilNode` as a condition

Separate issue:
`foo(())` (empty parens as a method argument) causes a different error:
`RuntimeError: *anonymous_rest`

This is unrelated to this PR, but it's worth addressing as a separate issue.

mame (approved):
Good catch! `DummyNilNode` has gradually been pulled into more places than I originally intended, so I should probably revisit its design at some point. That said, returning it here is much better than crashing, so I'll go ahead and merge this. Thanks for the fix!

PR #442, opened 2026-04-25
Support LSP 3.17 `positionEncoding` negotiation

ahogappa:
## Summary

  Implements LSP 3.17 `positionEncoding` negotiation. Today TypeProf hardcodes UTF-16LE columns and omits `positionEncoding` from its capabilities response, so modern clients that prefer UTF-8 (Helix, Zed, Neovim) all fall back to UTF-16.

  This PR also exposes the encoding as `Service.new(position_encoding:)` for embedders.
  The motivating use case is [ruby-minify](https://github.com/ahogappa/ruby-minify), which cross-references TypeProf nodes against Prism re-parses to drive variable/method renaming. 
Prism's native columns are byte-based — equivalent to LSP `utf-8` — so `Service.new(position_encoding: Encoding::UTF_8)` makes the two coordinate systems line up. 
Without this option, ruby-minify reaches into TypeProf via `node.instance_variable_get(:@raw_node).location` to bypass UTF-16 columns.

  ## Behavior change

  None by default. Clients that don't send `general.positionEncodings`, and embedders that don't pass `position_encoding`, still get UTF-16LE.

  ## Implementation note

  The server picks the first mutually-supported encoding from the client's preference-ordered list (`utf-8` / `utf-16` / `utf-32`), falling back to `utf-16` per spec.

  `FileContext#column_offsets_for` returns columns in the configured encoding. For UTF-8 it uses Prism's native `start_column` / `end_column` (= byte offsets) — note that **Prism's `code_units_cache(Encoding::UTF_8)` reports code points rather than bytes**, which would violate the LSP spec.

  ## Verification

  End-to-end against Helix, which proposes `general.positionEncodings: ["utf-8", "utf-32", "utf-16"]`. Source — line 4 references undefined `あfoo`, where `あ` is 3 bytes in UTF-8 / 1 code unit in UTF-16:

  ```ruby
  def foo(n)
    n
  end

  あfoo(1, 2)
  ```

  ### Helix `helix_lsp::transport` log — `InitializeResult`

  **Master (`ruby/typeprof` 0.31.1):**
  ```
  typeprof <- {"id":0,"result":{"capabilities":{"textDocumentSync":{...},"hoverProvider":true,...,"referencesProvider":true},"serverInfo":{"name":"typeprof","version":"0.31.1"}},"jsonrpc":"2.0"}
  ```

  **This PR:**
  ```
  typeprof <- {"id":0,"result":{"capabilities":{"positionEncoding":"utf-8","textDocumentSync":{...},"hoverProvider":true,...,"referencesProvider":true},"serverInfo":{"name":"typeprof","version":"0.31.1"}},"jsonrpc":"2.0"}
  ```

  The only difference is the new `"positionEncoding":"utf-8"` field, which signals that subsequent column values use UTF-8 byte offsets.

  - Unit tests in `test/core/position_encoding_test.rb` (UTF-8 / UTF-16 / UTF-32 columns).
  - LSP tests in `test/lsp/lsp_test.rb` (negotiation matrix).
  - End-to-end against Helix: confirmed `positionEncoding: "utf-8"` response and correct diagnostic range on a non-ASCII identifier.

  Spec: <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments>

### DIFF
--- a/lib/typeprof/core/env.rb
+++ b/lib/typeprof/core/env.rb
@@ -259,14 +259,14 @@ module TypeProf::Core
       mod.get_type_alias(name)
     end
 
-    def load_core_rbs(raw_decls)
-      file_context = FileContext.new(nil)
+    def load_core_rbs(raw_decls, position_encoding)
+      file_context = FileContext.new(nil, position_encoding)
       lenv = LocalEnv.new(file_context, CRef::Toplevel, {}, [])
       decls = raw_decls.map do |raw_decl|
         AST.create_rbs_decl(raw_decl, lenv)
       end.compact
 
-      decls += AST.parse_rbs("typeprof-rbs-shim.rbs", <<-RBS)
+      decls += AST.parse_rbs("typeprof-rbs-shim.rbs", <<-RBS, position_encoding)
         class Exception
           include _Exception
         end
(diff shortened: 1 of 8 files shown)


lib/typeprof/core/env.rb
```diff
@@ -308,16 +308,34 @@ class Object
   end
 
   class FileContext
-    attr_reader :path, :comments
-    def initialize(path, prism_source = nil, comments = nil)
+    attr_reader :path, :comments, :position_encoding
+    def initialize(path, position_encoding = nil, prism_source = nil, comments = nil)
       @path = path
+      @position_encoding = position_encoding || Encoding::UTF_16LE
```
ahogappa, line 314: FYI: given that Ruby's default source encoding is UTF-8, UTF-8 would arguably be a more natural default. 
However, switching from UTF-16LE would be a breaking change for existing users, so this PR keeps UTF-16LE.

mame (approved):
Thanks, I would like to give it a try!

PR #444, opened 2026-04-27
Support `class Foo = Bar` / `module Foo = Bar` declarations in RBS

rhiroe:
## Summary

Implement RBS class alias and module alias declarations (`class Foo = Bar`, `module Foo = Bar`).

Builds on #443, which fixed the `NoMethodError` caused by these declarations being silently skipped without `.compact`. With that crash gone, the alias was still being dropped — every `YAML.load` / `HashWithIndifferentAccess.foo` etc. came back untyped. This PR makes those resolve.

After this PR, `module YAML = Psych` makes the following resolve identically to `Psych`:

- singleton method call: `YAML.load(str)`
- nested constant: `YAML::SyntaxError`
- mixin: `class A; include YAML end`

This also unblocks real-world RBS sources such as `module YAML = Psych` (commonly authored by users), `module Sidekiq::ClientMiddleware = Sidekiq::ServerMiddleware` (gem_rbs_collection sidekiq), and `class HashWithIndifferentAccess = ActiveSupport::HashWithIndifferentAccess` (gem_rbs_collection activesupport).

## Approach

The change centered on `Genv#resolve_cpath`:

- `ModuleEntity` gains `@alias_decls` (decl ⇒ target_mod) and `@alias_target`, with `add_alias_decl` / `remove_alias_decl` mirroring the existing module decl lifecycle. The alias decl is also registered as a const on the outer module so const lookup of the alias name succeeds.
- `Genv#resolve_cpath` follows `alias_target` after resolving each cpath segment, with cycle detection to handle `module A = B; module B = A`.
- New AST nodes `SigClassAliasNode` / `SigModuleAliasNode` register the alias via the standard `define` / `undefine` / `install` lifecycle. They access the alias's own `ModuleEntity` via `outer.inner_modules[cname]` directly to avoid being short-circuited by `resolve_cpath`'s alias following.
- The previously empty `when RBS::AST::Declarations::AliasDecl` branch in `AST.create_rbs_decl` is replaced with explicit `when ClassAlias` / `when ModuleAlias` branches.

## Test plan

- [x] `scenario/rbs/class-module-alias.rb`: class alias instance method, module alias singleton method / nested constant / mixin
- [x] `scenario/rbs/class-module-alias-nested.rb`: alias declared inside a module (`module Outer; module InnerAlias = Inner; end`)
- [x] `scenario/rbs/class-module-alias-cycle.rb`: cycle (`module A = B; module B = A`) does not infinite-loop
- [x] `bundle exec rake test` — 417 tests, 772 assertions, 0 failures

## Known limitations / follow-ups

- LSP "go to definition" on an aliased name currently jumps to the target module's declaration site, not the alias declaration site. A follow-up PR will add the alias's own source location for LSP.

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -436,7 +436,10 @@ module TypeProf::Core
         SigInterfaceNode.new(raw_decl, lenv)
       when RBS::AST::Declarations::Constant
         SigConstNode.new(raw_decl, lenv)
-      when RBS::AST::Declarations::AliasDecl
+      when RBS::AST::Declarations::ClassAlias
+        SigClassAliasNode.new(raw_decl, lenv)
+      when RBS::AST::Declarations::ModuleAlias
+        SigModuleAliasNode.new(raw_decl, lenv)
       when RBS::AST::Declarations::TypeAlias
         SigTypeAliasNode.new(raw_decl, lenv)
         # TODO: check
--- a/lib/typeprof/core/ast/sig_decl.rb
+++ b/lib/typeprof/core/ast/sig_decl.rb
@@ -414,6 +414,52 @@ module TypeProf::Core
       end
     end
 
+    class SigModuleAliasBaseNode < Node
+      def initialize(raw_decl, lenv)
+        super(raw_decl, lenv)
+        @cpath = AST.resolve_rbs_name(raw_decl.new_name, lenv)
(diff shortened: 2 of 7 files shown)


mame:
@rhiroe Thanks! The PR looks good to me. Is it still marked as Draft because you're planning to do more work on it? If not, I'll go ahead and merge.

rhiroe:
@mame I've marked this PR as ready for review.

mame:
Thanks!

PR #447, opened 2026-05-03
Fix NoMethodError in CallBaseNode#modified_vars for anonymous rest

sinsoku:
Since ff8a5167, anonymous rest arguments (bare `*`) are represented as
`nil` placeholders in `@positional_args` instead of `DummyNilNode`.
However, `CallBaseNode#modified_vars` iterates the positional args
array without skipping nil, raising NoMethodError when a call with
anonymous rest forwarding appears inside a branch (e.g.
`if cond; bar(*); end`) whose `modified_vars` is walked from
`BranchNode#install0`.

Use safe navigation so nil placeholders are skipped, matching how
`each_subnode` already guards against nil subnodes. Add a regression
case to scenario/args/anonymous_rest.rb.

### DIFF
--- a/lib/typeprof/core/ast/call.rb
+++ b/lib/typeprof/core/ast/call.rb
@@ -247,7 +247,7 @@ module TypeProf::Core
               subnode.modified_vars(tbl, vars)
             end
           else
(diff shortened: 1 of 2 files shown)


mame (approved):
:+1:

PR #449, opened 2026-05-05
Add a benchmark workflow for real-world projects

sinsoku:
## Motivation / Background

The scenario tests only cover small inputs, so nothing tells us when a change slows the analysis down, drops type coverage, or crashes on a real-world codebase.

## Detail

`tool/benchmark.rb` analyzes four pinned projects — TypeProf itself (v0.32.0), Optcarrot, RubyGems.org and Redmine (7.0.1) — and reports the analysis time and the type coverage per slot.

- **`.github/workflows/benchmark.yml`** (new) runs it on every push to master and publishes both metrics with [benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark), which keeps the history as a chart.
- **`.github/workflows/main.yml`** runs it on pull requests, where nothing is published and only a crash or a hang fails the job. It uses the latest released Ruby only, since a crash shows up on any version. This replaces the `Run the dog bench` step; `tool/dog_bench.rb` stays for ad-hoc profiling.

CI cost: in a run on my fork the step took 81s on `build (4.0)`, nearly all of it the analysis itself. Since the other matrix entries skip it, the workflow goes from about 40 seconds to about two minutes.

Example (running on my fork): https://sinsoku.github.io/typeprof/dev/bench/

## Setup

Two separate things are needed before the charts show up.

- [x] **A `gh-pages` branch** — done. I created an empty one on this repository:

  ```sh
  git switch --orphan gh-pages
  git commit --allow-empty -m "Initialize gh-pages"
  git push origin gh-pages
  git switch master
  ```

- [x] **GitHub Pages enabled** — not done, and only an admin can do it. Settings → Pages → Source: `Deploy from a branch`, `gh-pages` / (root).

The workflow stores the data either way, so the second item does not block merging. Without it the dashboard is simply not visible.

## Resolved questions

1. ~~Remove `tool/dog_bench.rb`?~~ -> **Kept**, as noted above.
2. ~~Run benchmarks on each PR?~~ -> **Added as a test, not a benchmark**, as you suggested.


sinsoku:
@mame 
What do you think of the overall direction? Want to make sure I'm on the right track.

mame:
This looks really nice! I'd like to give it a try.

A few thoughts on the open questions:

1. `tool/dog_bench.rb`: I'd like to keep it for now. I do use it for ad-hoc profiling sometimes.
2. Running on each PR: I think this would be valuable, but more as a test than as a benchmark. It would help us catch cases where a change to TypeProf accidentally breaks it on Redmine or Optcarrot.

sinsoku:
@mame 
I've resolved both points.

1. Kept `tool/dog_bench.rb` as-is.
2. Run the benchmark scripts in CI.

The CI uses `tool/benchmarks/typeprof.rb` instead of `tool/dog_bench.rb`, because pinning the analyzed version keeps the input stable and makes regressions easier to detect.

sinsoku:
@mame 
I rewrote this with a different approach, so the diff is entirely new.

It now uses [benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) instead of the dashboard I had hand-rolled, so there is much less of our own code to maintain. I also dropped the per-project setup that ran `bundle add`, `rbs collection install` and `rbs_rails`, since it made CI depend on resolving gems at run time and could fail non-deterministically. The numbers now reflect what TypeProf infers from the projects as they are, with no RBS set up for them.

I have updated the description, so please see there for the details. Could you take another look when you have time?

mame (approved):
Thank you!

PR #452, opened 2026-05-08
Support Symbol#to_proc block inference

### LINKED ISSUE #414: Symbol#to_proc pattern (&:method_name) is not inferred correctly
## Problem
```ruby
["1", "2"].map(&:to_i)         # => Array[untyped]  (wrong)
["1", "2"].map { |x| x.to_i }  # => Array[Integer]  (correct)
```

The `Symbol#to_proc` pattern is currently not inferred correctly.

## Root Cause

`&:to_i` is treated as `Type::Symbol` in `a_args.block`, but `MethodDeclBox#resolve_overload` (`graph/box.rb`) only handles `Type::Proc`. As a result, `Type::Symbol` is silently ignored, and the return type remains `untyped`.

## Proposed Fix

Add a `when Type::Symbol` branch in `MethodDeclBox#resolve_overload`:
```ruby
when Type::Symbol
  recv_vtx = blk_a_args[0]  # element type (e.g. String)
  call_box = changes.add_method_call_box(genv, recv_vtx, ty.sym, ActualArguments.new([], [], nil, nil), false)
  changes.add_edge(genv, call_box.ret, ret_vtx)
```

This works generically for any symbol (`:to_i`, `:upcase`, etc.) and any block-accepting method (`map`, `select`, etc.) without touching `builtin.rb`.

## Discussion

Conceptually, `Symbol#to_proc` handling is very similar to `Kernel#send` (dynamic method resolution from a Symbol), which is already handled in `builtin.rb`. Therefore, I believe that `builtin.rb` is the most natural and "principled" place for this fix.

However, I recall a comment in [PR #410](https://github.com/ruby/typeprof/pull/410) where it was mentioned there is no intention to keep adding features to `builtin.rb`. For this reason, I am proposing the fix in `graph/box.rb` as an alternative.

I couldn't find any existing `TODO`s related to `Symbol#to_proc`, so I am also wondering if this current behavior (returning `untyped`) is an intentional design choice or simply not prioritized yet.

Any feedback on whether this is an issue you'd accept a PR for, and if so, your preferred approach, would be greatly appreciated!

pvcresin:
## Summary

Fixes #414 

Support `Symbol#to_proc` blocks such as `map(&:to_i)`.

This treats a symbol passed as a block as a method call where the first yielded block argument is the receiver and the remaining yielded arguments are method arguments.

## Changes

- Infer `["1", "2"].map(&:to_i)` as `Array[Integer]`
- Support the same Symbol#to_proc behavior for Ruby `yield`
- Add scenario coverage for RBS block calls and Ruby method block calls

### DIFF
--- a/lib/typeprof/core/graph/box.rb
+++ b/lib/typeprof/core/graph/box.rb
@@ -43,6 +43,15 @@ module TypeProf::Core
       raise NotImplementedError
     end
 
+    def add_symbol_proc_call_box(changes, genv, sym, caller_positionals)
+      return if caller_positionals.empty?
+
+      recv = caller_positionals.first
+      positionals = caller_positionals[1..]
+      a_args = ActualArguments.new(positionals, ::Array.new(positionals.size, false), nil, nil)
+      changes.add_method_call_box(genv, recv, sym, a_args, false)
+    end
+
     def to_s
       "#{ self.class.to_s.split("::").last }#{ @id ||= $new_id += 1 }"
     end
@@ -382,6 +391,8 @@ module TypeProf::Core
                 end
               end
             end
+          when Type::Symbol
+            resolve_symbol_proc(changes, genv, ty.sym, blk_a_args, rbs_blk, param_map0)
           end
         end
       end
@@ -394,6 +405,13 @@ module TypeProf::Core
         ret_vtx = method_type.return_type.covariant_vertex(genv, changes, param_map0)
         changes.add_edge(genv, ret_vtx, ret)
       end
+    end
+
+    def resolve_symbol_proc(changes, genv, sym, blk_a_args, rbs_blk, param_map)
+      box = add_symbol_proc_call_box(changes, genv, sym, blk_a_args)
+      return unless box
+
+      rbs_blk.return_type.typecheck(genv, changes, box.ret, param_map)
     end
 
     def resolve_overloads(changes, genv, node, param_map, a_args, ret, &blk)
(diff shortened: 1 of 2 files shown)


sinsoku:
I tried this branch against Redmine 6.1.1 with [this script](https://gist.github.com/sinsoku/bff00848fdd341512ac3edbd9f07c69a), and the analysis hangs (master finishes in under 20s). Could you take a look?

lib/typeprof/core/graph/box.rb
```diff
@@ -1026,7 +1044,10 @@ def run0(genv, changes)
       called_mdefs = Set.empty
       error_count = 0
       resolve(genv, changes) do |me, ty, mid, orig_ty|
-        if !me
+        if @node.is_a?(AST::YieldNode) && mid == :call && orig_ty.is_a?(Type::Symbol)
+          box = add_symbol_proc_call_box(changes, genv, orig_ty.sym, @a_args.positionals)
```
sinsoku: `yield` with keyword arguments is valid Ruby, e.g.:

```ruby
def foo
  yield 1, key: "v"
end
foo { |x, key:| puts "x=#{x}, key=#{key}" } # => x=1, key=v
```

So I think `@a_args.keywords` should also be passed through to `add_symbol_proc_call_box` here.

pvcresin: Good catch — fixed in 7b8e9b0f. `@a_args.keywords` is now forwarded, with a kwarg-yield scenario added.

pvcresin:
@sinsoku Reproduced the Redmine 6.1.1 hang locally. 
Each parent re-run allocated a fresh sub-box, so `box.ret` had a new identity every cycle and the typecheck edge churned the parent into an infinite re-run. 
Fixed in 70adf42c by memoizing the sub-box on the parent — Redmine now finishes in ~5s, matching master.

sinsoku (approved):
Thanks!

PR #448, opened 2026-05-04
Add release workflow to automate gem publishing

sinsoku:
## Motivation / Background

I want to increase the release frequency by automating the gem release process.

## Detail

Allow publishing a new version to RubyGems simply by merging a release PR.
[tagpr](https://github.com/Songmu/tagpr) handles release PR creation and tagging, and [rubygems/release-gem](https://github.com/rubygems/release-gem) pushes the gem.

## Required setup (cc: @mame)

Several settings are required for this workflow to work.

- [ ] Enable "Allow GitHub Actions to create and approve pull requests"
    - [in repository settings](https://github.com/ruby/typeprof/settings/actions) (Settings → Actions → General → Workflow permissions)
    - tagpr needs this to open the release PR
- [x] Configure RubyGems Trusted Publisher
    - [in typeprof settings](https://rubygems.org/gems/typeprof/trusted_publishers)
    - Repository owner: `ruby`
    - Repository name: `typeprof`
    - Workflow filename: `release.yml`
    - Environment: `release`
    - refs: https://guides.rubygems.org/trusted-publishing/

## How the release flow works

1. When a commit is pushed to `master`, tagpr opens (or updates) a release PR
   titled something like `Release for vX.Y.Z`. The PR contains:
   - A version bump in `lib/typeprof/version.rb`
   - An updated `Gemfile.lock` (via `postVersionCommand`)
   - An auto-generated `CHANGELOG.md` entry summarizing merged PRs since
     the last release
2. The next version defaults to a patch bump. To bump minor/major, use either:
   - **Label a merged PR** with `minor` or `major` — tagpr will automatically
     propagate this to the release PR as `tagpr:minor` / `tagpr:major`, or
   - **Label the release PR directly** with `tagpr:minor` or `tagpr:major`
     to override / set the bump manually.
3. Merging the release PR causes tagpr to push a `vX.Y.Z` tag and create a
   GitHub Release.
4. The tag triggers `rubygems/release-gem`, which builds and pushes the gem
   to RubyGems via Trusted Publishing (no API key needed).

## Examples                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                      
* https://github.com/Songmu/tagpr/pull/347                                                                                                                                                                                                                                                            
* https://github.com/sinsoku/ruby-lsp-typeprof/pull/3

PR #453, opened 2026-05-09
Set up release automation via Trusted Publisher

sinsoku:
## Motivation / Background

Automate the gem release process so that maintainers only need to
land a version bump commit on master.

## How the release flow works

1. Maintainer runs `ruby tool/bump.rb` to create a version bump
   commit on master.
2. The release workflow detects the change to
   `lib/typeprof/version.rb`, pushes the `vX.Y.Z` tag, publishes
   the gem via Trusted Publishing, and creates a GitHub Release
   with auto-generated notes.

## Required setup

- [x] RubyGems Trusted Publisher is configured (via #448, with
  `release.yml` and `release` environment)

## Notes

- This PR does not introduce a `CHANGELOG.md`. Release notes are
  produced via `gh release create --generate-notes`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,35 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - lib/typeprof/version.rb
+
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: write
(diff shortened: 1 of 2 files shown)


sinsoku:
@mame 
I think using a script to simply increment the version and a trusted publisher would be a better fit — it avoids the org-level setting change.

What do you think?

mame (approved):
Go ahead!

PR #455, opened 2026-05-10
Use bundled Bundler on ruby-head in CI

sinsoku:
Released Bundler is broken on ruby-head due to ruby/ruby#16895.
The fix (rubygems/rubygems#9529) has not been released yet.


mame (approved):
👏

PR #461, opened 2026-06-16
Merge duplicate constant declarations into a union type

pvcresin:
When a constant is assigned more than once, `Service#dump_declarations`
emitted one declaration line per assignment, so

```ruby
class C
  D = 1
  D = "str"
end
```

was dumped as two declarations for the same constant:

```rbs
class C
  D: Integer
  D: String
end
```

This deduplicates constant writes by their static cpath so that each
constant yields a single declaration, and shows the union of its types
when it is assigned more than once:

```rbs
class C
  D: (Integer | String)
end
```

A single assignment keeps using the assigned value (`node.ret`) rather
than the constant entity's type, so that a same-named class or module
(e.g. one declared in RBS) is not mixed into the declaration; see
`scenario/rbs/remove-class.rb`, where `C = 1` coexists with an RBS
`class C` and the declaration is expected to stay `C: Integer`. The
deduplication mirrors the existing `seen_ivars`/`seen_cvars`/`seen_gvars`
handling in the same method.

`scenario/known-issues/multi-const-write.rb` now passes and is promoted
to `scenario/const/`. `scenario/variable/operator_write.rb` is updated
since it previously encoded the duplicated output (`D = 0; D += 1` and
`C::E = 0; C::E += 1` now each produce a single line).

All 429 tests pass (`bundle exec rake test`).


sinsoku (approved):
It looks good!

PR #462, opened 2026-06-21
Support `extend`

sinsoku:
TypeProf did not handle `extend`, so methods added via `extend M` were reported as "undefined method" and `extend` was missing from the dumped RBS.

Implement it symmetrically to `include`: parse Ruby/RBS `extend`, resolve the extended module's instance methods as the receiver's singleton methods, and dump `extend` for class bodies (like `include`).

Self-referential `extend` (e.g. `extend self`) is still unsupported.

### DIFF
--- a/lib/typeprof/core/ast.rb
+++ b/lib/typeprof/core/ast.rb
@@ -283,6 +283,8 @@ module TypeProf::Core
           case raw_node.name
           when :include
             return IncludeMetaNode.new(raw_node, lenv)
+          when :extend
+            return ExtendMetaNode.new(raw_node, lenv)
           when :attr_reader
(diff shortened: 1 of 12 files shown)


mame:
Thanks, merged! The symmetry with `include` looks right to me.

One thing I noticed while testing: `include` and `extend` can form a cycle that is valid Ruby, and it makes `on_ancestors_updated` recurse forever.

```ruby
module Foo
  module Util
    include Foo
    def helper = "h"
  end
  extend Util
end
Foo.helper
```

The guard was already missing before this PR, but `include`/`prepend` cycles are rejected by Ruby itself, so this combination is the only way to reach it. I'll push a follow-up with a one-line guard and a scenario.

PR #464, opened 2026-06-23
Support type inference for pattern-matching capture variables

pvcresin:
## Summary

Capture variables in `case/in` patterns were always inferred as `untyped` because the matched value was never threaded into the patterns:

```ruby
case a
in Integer => n
  n + 1        # n was untyped
in [String => s, Symbol => t]
  [s, t]       # s, t were untyped
end
```

## Changes

- Add an `install_pattern` / `install_pattern0` pair (mirroring `install` / `install0`, so `@changes` is reconciled the same way during incremental analysis) that flows the matched value into each pattern node.
- Local targets bind to the matched value; `Const => var` narrows it by the class the same way `when Const` does (reusing `IsAFilter`); array / find patterns decompose it into element vertices via splat boxes. `CaseMatchNode` and the one-line `=>` / `in` forms pass the subject in.
- Hash patterns still capture as `untyped` (extracting a value type per key needs a dedicated box); sub-patterns inside them are still matched.
- Promote `scenario/known-issues/pattern-capture-narrowing.rb` into `scenario/patterns/capture.rb` and update `scenario/patterns/var_pat.rb`.

All 433 tests pass (`bundle exec rake test`).


mame (approved):
Took me a while to get my head around it, but this looks good!

PR #465, opened 2026-08-10
Fix crash on anonymous splat in find pattern and `{**nil}`

ahogappa:
Two pattern-matching forms crash during AST construction on master:

```ruby
case ary
in [*, x, *rest]
  x
end
# => lib/typeprof/core/ast.rb:334:in 'create_pattern_node':
#    undefined method 'type' for nil (NoMethodError)
```

```ruby
case h
in {**nil}
  1
end
# => lib/typeprof/core/ast/pattern.rb:42:in 'initialize':
#    undefined method 'value' for an instance of Prism::NoKeywordsParameterNode
```

Prism represents an anonymous splat as a `SplatNode` whose `expression` is `nil`, and `HashPatternNode#rest` can be a `NoKeywordsParameterNode`, which has no `value`. Both were passed to `AST.create_pattern_node` unchecked. `ArrayPatternNode` already guards against these, so this applies the same handling to `FindPatternNode` and `HashPatternNode`.

### DIFF
--- a/lib/typeprof/core/ast/pattern.rb
+++ b/lib/typeprof/core/ast/pattern.rb
@@ -58,9 +58,9 @@ module TypeProf::Core
     class FindPatternNode < Node
       def initialize(raw_node, lenv)
         super(raw_node, lenv)
-        @left = raw_node.left ? AST.create_pattern_node(raw_node.left.expression, lenv) : nil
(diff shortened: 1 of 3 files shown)


sinsoku (approved):
Thanks!

PR #467, opened 2026-08-17
Fix splat fallback for union types

pvcresin:
## Summary

- Fix splat fallback for union types.
- Move the regression scenario into the regular array scenarios.

## Details

`[*x]` now keeps receiver types without `to_a` separate from the normal `to_a` result, so each union member can be handled independently. Existing array and tuple inference remains unchanged.

## Test

- `bundle exec ruby -Itest test/scenario_test.rb`
- `bundle exec rake test`


mame:
Thanks! `fallback` feels like the name from the splat's side; I'd prefer something like `unresolved_recv`. It's trivial, so let me merge this as is and rename in a follow-up.

PR #477, opened 2026-09-08
Resolve type variables of a reopened declaration by position

sinsoku:
## Motivation

RBS 4.1 renamed the core's type parameters (`Array[Elem]` became `Array[E]`), but RBS files in rbs collections still reopen them with the old names: [activesupport in gem_rbs_collection](https://github.com/ruby/gem_rbs_collection/blob/3913b7dffb65bdffdfe5cb9095108f932e85433b/gems/activesupport/7.0/activesupport-7.0.rbs#L97) has `module Enumerable[unchecked out Elem]`, and so does the rbs gem's own [sig/shims/enumerable.rbs](https://github.com/ruby/rbs/blob/v4.2.0/sig/shims/enumerable.rbs). RBS allows this, as `validate_type_params` compares the parameters after renaming them.

TypeProf keyed type variable substitution by the names of a single declaration, so with such a collection any method lookup through `Enumerable` crashed and the whole file failed to analyze:

```
$ bin/typeprof lib/typeprof/cli/cli.rb
lib/typeprof/core/ast/sig_type.rb:887:in 'TypeProf::Core::AST::SigTyVarNode#covariant_vertex0': unknown type variable: Elem (RuntimeError)
        from lib/typeprof/core/env.rb:142:in 'block in TypeProf::Core::GlobalEnv#get_instance_type'
        from lib/typeprof/core/graph/box.rb:1269:in 'block in TypeProf::Core::MethodCallBox#resolve_included_modules'
```

#468 met the same limitation with TypeProf's own shim and aligned the shim's names with the core (91bd108a), which is not an option for third-party RBS.

## Changes

`SigTyVarNode` maps a name missing from the substitution to the entity's name at the same position in its own declaration. This is done at lookup time instead of renaming the AST, because the entity's names can change when a declaration is removed in the LSP, while the position stays the same. The workaround from #468 is reverted, so the shim exercises the mapping every time the core RBS is loaded.

## Verification

- CI passes.
- `bin/typeprof lib` with an `rbs_collection.yaml` that includes the rbs gem completes without errors.
- The new scenarios fail before the fix with `unknown type variable: X` / `T`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/typeprof/core/ast/sig_type.rb
```diff
@@ -883,18 +883,34 @@ def initialize(raw_decl, lenv)
...
+      # A reopened declaration may rename the type parameters (e.g., `Enumerable[E]`
+      # and `Enumerable[Elem]`), but the subst is keyed by the names of the module
+      # entity, so fall back to matching them by position
+      def resolve_var(genv, subst)
+        var_vtx = subst[@var]
+        return var_vtx if var_vtx
+        decl_cpath, decl_params = @lenv.sig_type_params
+        return nil unless decl_params
+        idx = decl_params.index(@var)
+        return nil unless idx
```
mame:
```suggestion
      def resolve_var(genv, subst)
        decl_cpath, decl_params = @lenv.sig_type_params
        return subst[@var] unless decl_params
        idx = decl_params.index(@var)
        return subst[@var] unless idx
```

mame: One suggestion: resolve by position only, rather than trying the name first.

A method type parameter is not in the declaration's parameter list, so it still goes through the name. This also gets `class Hash[V, K]; def key_of: () -> V` right (`:a`, not `Integer`). The suite passes either way.

mame (dismissed):
Thanks! Just one place I'd like to simplify. If this looks good to you, feel free
to merge it yourself.

